PR #6, opened 2024-08-26
Add ND version and site resources (DCNE-179)

sajagana:


samiib (changes requested):
I can see the CI is failing.
To get past that step I think you need to run chmod to change the `scripts/gofmtcheck.sh` permissions to 755. 
I think something like `git update-index --chmod=+x scripts/gofmtcheck.sh` would do the trick.

internal/provider/resource_nd_site.go
```diff
@@ -0,0 +1,491 @@
...
+
+			if attributeName == "latitude" {
+				data.Latitude = basetypes.NewStringValue(attributeValue.(string))
+			}
+
+			if attributeName == "longitude" {
+				data.Longitude = basetypes.NewStringValue(attributeValue.(string))
+			}
+
+			if os.Getenv("ND_LOGIN_DOMAIN") != "" {
```
shrsr, line 481: Does API return the correct value for login_domain when env var "ND_LOGIN_DOMAIN" is set?

internal/provider/resource_nd_site.go
```diff
@@ -0,0 +1,491 @@
...
+func (r *SiteResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	tflog.Debug(ctx, "Start import state of resource: nd_site")
+	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
+
+	var stateData *SiteResourceModel
+	resp.Diagnostics.Append(resp.State.Get(ctx, &stateData)...)
+
+	username := os.Getenv("ND_SITE_USERNAME")
+	if username == "" {
+		resp.Diagnostics.AddError("Missing input", "A username must be provided during import, please set the ND_SITE_USERNAME environment variable")
```
shrsr, line 334: ->
"The username of the ND site must be provided during import, please set the ND_SITE_USERNAME environment variable"

internal/provider/resource_nd_site.go
```diff
@@ -0,0 +1,491 @@
...
+	}
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("username"), username)...)
+
+	password := os.Getenv("ND_SITE_PASSWORD")
+	if password == "" {
+		resp.Diagnostics.AddError("Missing input", "A password must be provided during import, please set the ND_SITE_PASSWORD environment variable")
```
shrsr, line 340: ->
"The password of the ND site must be provided during import, please set the ND_SITE_PASSWORD environment variable"

internal/provider/resource_nd_site.go
```diff
@@ -0,0 +1,491 @@
...
+	}
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("password"), password)...)
+
+	loginDomain := os.Getenv("ND_LOGIN_DOMAIN")
+	if loginDomain == "" {
+		resp.Diagnostics.AddError("Missing input", "A login_domain must be provided during import, please set the ND_LOGIN_DOMAIN environment variable")
```
shrsr, line 346: ->
"The login_domain of the ND site must be provided during import, please set the ND_LOGIN_DOMAIN environment variable"

internal/provider/resource_nd_site.go
```diff
@@ -0,0 +1,491 @@
...
+				data.Latitude = basetypes.NewStringValue(attributeValue.(string))
+			}
+
+			if attributeName == "longitude" {
+				data.Longitude = basetypes.NewStringValue(attributeValue.(string))
+			}
+
+			if os.Getenv("ND_LOGIN_DOMAIN") != "" {
+				data.LoginDomain = basetypes.NewStringValue(os.Getenv("ND_LOGIN_DOMAIN"))
+			} else if attributeName == "loginDomain" {
```
shrsr, line 483: This part of the code is not required anymore because data.LoginDomain was already set using getBaseSiteResourceModel
```
else if attributeName == "loginDomain" {
				data.LoginDomain = basetypes.NewStringValue(attributeValue.(string))
			}
```

shrsr (changes requested):

internal/provider/resource_nd_site.go
sajagana: API will not return login_domain. Because of that, the ND_LOGIN_DOMAIN must be set during the import.

shrsr: This still might be an issue because Terraform will likely complain when we the following scenario:
If both login_domain of the resource attribute and the env variable ND_LOGIN_DOMAIN have different values
We need to somehow isolate this **if os.Getenv("ND_LOGIN_DOMAIN") != "" {** for the import case

shrsr (changes requested):

akinross (approved):
LGTM

shrsr (dismissed):
LGTM

samiib (dismissed):
LGTM

gmicol (dismissed):
LGTM

anvitha-jain (approved):
LGTM

internal/provider/provider.go
```diff
@@ -0,0 +1,225 @@
...
+
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	username := getStringAttribute(data.Username, "ND_USERNAME")
+	password := getStringAttribute(data.Password, "ND_PASSWORD")
+	isInsecure := getBoolAttribute(resp, data.IsInsecure, "ND_INSECURE", true)
```
lhercot, line 140: This should not be defaulted to True. Why are we defaulting it to True?

internal/provider/resource_nd_site.go
```diff
@@ -0,0 +1,489 @@
...
+	password := os.Getenv("ND_SITE_PASSWORD")
+	if password == "" {
+		resp.Diagnostics.AddError("Missing input", "The password of the ND site must be provided during import, please set the ND_SITE_PASSWORD environment variable")
+	}
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("password"), password)...)
+
+	loginDomain := os.Getenv("ND_SITE_LOGIN_DOMAIN")
+	if loginDomain == "" {
+		resp.Diagnostics.AddError("Missing input", "The login_domain of the ND site must be provided during import, please set the ND_SITE_LOGIN_DOMAIN environment variable")
+	}
```
lhercot, line 349: Do we need this? You ignore it anyway after the create so why do we actually need it? Can't we use a ModifyPlan to take care of it at the next plan?

internal/provider/client.go
```diff
@@ -0,0 +1,454 @@
...
+const DefaultBackoffMinDelay int = 4
+const DefaultBackoffMaxDelay int = 60
+const DefaultBackoffDelayFactor float64 = 3
+
+// Client is the main entry point
+type Client struct {
+	BaseURL            *url.URL
+	httpClient         *http.Client
+	AuthToken          *Auth
+	Mutex              sync.Mutex
```
lhercot, line 45: Do we need BaseURL, AuthToken and Mutex to be externally accessible? should we make them start with a lower case character?

internal/provider/client.go
```diff
@@ -0,0 +1,454 @@
+package provider
```
lhercot, line 1: Should we put the client and auth files in a client package?

docs/resources/site.md
```diff
@@ -0,0 +1,83 @@
...
+* `type` (siteType) - (String) The type of the site.
+  * Valid Values: `aci`, `dcnm`, `third_party`, `cloud_aci`, `dcnm_ng`, `ndfc`.
+
+### Optional ###
+
+* `login_domain` (loginDomain) - (String) The login domain of the site.
+* `inband_epg` (inband_epg) - (String) The In-Band Endpoint Group (EPG) used to connect ND to the site.
+* `latitude` (latitude) - (String) The latitude location of the site.
+* `longitude` (longitude) - (String) The longitude location of the site.
+
```
lhercot, line 63: I think we should add security domain in this one too.

docs/index.md
```diff
@@ -0,0 +1,91 @@
...
+page_title: "Provider: ND"
+sidebar_current: "docs-nd-index"
+description: |-
+  The Cisco ND provider is used to interact with the resources provided by Cisco Nexus Dashboard.
+  The provider needs to be configured with the proper credentials before it can be used.
+---
+
+# Nexus Dashboard (ND)
+
+Cisco Nexus Dashboard is a central management console for multiple data center sites and a common platform for hosting Cisco data center operation services, such as Nexus Dashboard Insights and Nexus Dashboard Orchestrator. These services are available for all the data center sites and provide real time analytics, visibility, assurance for network policies and operations, as well as policy orchestration for the data center fabrics, such as Cisco ACI or Cisco NDFC.
```
lhercot, line 12:
```suggestion
Cisco Nexus Dashboard is a central management console for multiple data center sites and a common analytics solution for Cisco data center operations. Nexus Dashboard allows users to manage multiple data center sites and provide real time analytics, visibility, assurance for network policies and operations, as well as policy orchestration for the data center fabrics, such as Cisco ACI or Cisco NDFC or standalone Nexus 9000 switches.
```

docs/index.md
```diff
@@ -0,0 +1,91 @@
...
+
+# Cisco ND Provider
+
+The Cisco ND terraform provider is used to interact with resources provided by Cisco Nexus Dashboard. The provider needs to be configured with proper credentials to authenticate with Cisco Nexus Dashboard.
+
+## Authentication
+
+Authentication with user-id and password.
```
lhercot, line 20:
```suggestion
Authentication with username and password.
```

docs/index.md
```diff
@@ -0,0 +1,91 @@
...
+```hcl
+provider "nd" {
+  username     = "admin"
+  password     = "password"
+  url          = "https://my-cisco-nd.com"
+  login_domain = "DefaultAuth"
+}
+```
+
+In this method, it will obtain an authentication token from Cisco Nexus Dashboard and will use that token to authenticate. A limitation with this approach is Nexus Dashboard counts the request to authenticate and threshold it to avoid DOS attack. After too many attempts this authentication method may fail as the threshold will be exceeded. To avoid the above-mentioned problem Cisco Nexus Dashboard supports signature-based authentication.
```
lhercot, line 33: Are you sure about this?

docs/index.md
```diff
@@ -0,0 +1,91 @@
...
+- `url` (String) URL of the Cisco Nexus Dashboard web interface.
+  - Environment variable: `ND_URL`
+
+## Optional
+
+- `login_domain` (String) Login domain for the Nexus Dashboard Account.
+  - Default: `DefaultAuth`
+  - Environment variable: `ND_LOGIN_DOMAIN`
+- `insecure` (Boolean) Allow insecure HTTPS client.
+  - Default: `true`
```
lhercot, line 83: I don't think it should be insecure by default.

lhercot (changes requested):

internal/provider/client.go
anvitha-jain: @sajagana can you please elaborate on if we need BaseURL, AuthToken and Mutex to be externally accessible?

sajagana: @anvitha-jain Currently, we do not access the BaseURL, Mutex, and AuthToken outside the client. We can change it if required.

docs/resources/site.md
```diff
@@ -0,0 +1,92 @@
...
+Starting in Terraform version 1.5, an existing Site can be imported using [import blocks](https://developer.hashicorp.com/terraform/language/import) via the following configuration:
+
+```
+import {
+  name = "{name}"
+  to   = nd_site.example
+}
+```
+
+~> The values for `username`, `password`, and `login_domain` attributes will not be imported when the nd_site resource imports an already registered site from the Nexus Dashboard. Use the `-replace` option to change the `username`, `password`, and `login_domain` attributes for the imported site.
```
lhercot, line 88:
```suggestion
~> The values for `username`, `password`, and `login_domain` attributes will not be imported when the `nd_site` resource imports an already registered site from Nexus Dashboard. Modifying the `username`, `password`, and `login_domain` will not update the imported site configuration on Nexus Dashboard. Use the `-replace` option to force the site recreation and use the new provided `username`, `password`, and `login_domain` attributes for the imported site.
```

lhercot (changes requested):

akinross (approved):
LGTM

shrsr (approved):
LGTM

samiib (approved):
LGTM

gmicol (approved):
LGTM

anvitha-jain (approved):
LGTM

lhercot (approved):
LGTM

PR #19, opened 2026-04-09
Code  restructuring and new resources for ND 4.2  (NDA-25)

mdmohan:
This is the first commit for 4.2 ND unified provider.
Existing code in this repo is cleaned up for development of the unified provider
- vendored code is removed - shall use just go.mod/sum to track
- The post PR acceptance tests is removed as the pipeline doesn't exist now
- cleaned up the existing resources and datasources


akabhask (approved):

samiib:
@mdmohan Would it be okay to separate the vendor removal into its own PR? It would make this PR much quicker to load and review.

As discussed in Webex, we will review individual commits.

go.mod
```diff
@@ -1,18 +1,21 @@
-module github.com/CiscoDevNet/terraform-provider-nd
```
samiib, line 1: Should this remain as "github.com/CiscoDevNet/terraform-provider-nd"?
All our other providers use this pattern.

.github/workflows/checks.yml
```diff
@@ -22,9 +22,6 @@ jobs:
       - name: gofmt Check
         run: ./scripts/gofmtcheck.sh
       - run: go mod tidy
-      - run: go mod vendor
-      - name: Check vendor for changes
-        run: git diff --exit-code
```
samiib, line 27: Can we keep the diff check just on go.mod and go.sum? eg.
```suggestion
      - name: Check go.mod and go.sum for changes
        run: git diff --exit-code go.mod go.sum
```

internal/manage/api/fabric_api.go
```diff
@@ -0,0 +1,103 @@
...
+	papi := new(FabricAPI)
+	papi.mutex = lock
+	papi.Client = client
+	papi.NDManageAPI = papi
+	return papi
+}
+
+func (c *FabricAPI) GetLock() *sync.Mutex {
+	return c.mutex
+}
```
akinross, line 51: Why is this method defined at the specific implementation level when it is the same for every api? is the mutex and GetLock()  generic for every implementation, where they return their own mutex status?

internal/manage/api/fabric_api.go
```diff
@@ -0,0 +1,103 @@
...
+
+// FabricAPI is the API client for fabric resources
+const (
+	// Standard Fabric API endpoints
+	UrlFabrics      = "/manage/fabrics"
+	UrlFabricByName = "/manage/fabrics/%s"
+
+	// Legacy endpoints - keeping for compatibility
+	UrlFabricGetAll         = "/lan-fabric/rest/control/fabrics"
+	urlPerFabric            = UrlFabricGetAll + "/%s"
```
akinross, line 26: is there a reason to mix camel and pascal casing for variable naming?

internal/manage/api/api.go
```diff
@@ -0,0 +1,156 @@
...
+
+type NDManageAPI interface {
+	GetLock() *sync.Mutex
+	//ProcessResponse(ctx context.Context, res gjson.Result) ([]string, error)
+	GetUrl() string
+	PostUrl() string
+	PutUrl() string
+	DeleteUrl() string
+	GetDeleteQP() []string
+	RscName() string
```
akinross, line 30: Would it be an option to use less abbreviations in the attribute naming, or use comments? For me it is hard to quickly understand what QP and Rsc stands for

docs/resources/fabric_vxlan.md
```diff
@@ -0,0 +1,516 @@
+---
```
akinross, line 1: we need to track cleanup of docs prior to release somewhere, is there any tracking mechanism in place already for these kind of activities?

internal/manage/api/config_api.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) 2024 Cisco Systems, Inc. and its affiliates
```
akinross: assume it is left but not used for now?

internal/common/types/custtypes.go
```diff
@@ -0,0 +1,12 @@
+// Code generated;  DO NOT EDIT.
```
akinross, line 1: just generic question what is the difference between the types and the custtypes file? would it be helpful to write a comment on the intent of these constructs?

should we use to snakecase for custtypes to keep consistent with other file names something like: cust_types, custom, custom_types ?

internal/manage/api/inventory_api.go
```diff
@@ -0,0 +1,164 @@
...
+func (c *InventoryAPI) WithFabric(fabricName string) *InventoryAPI {
+	c.FabricName = fabricName
+	return c
+}
+
+// WithSerial sets the serial number
+func (c *InventoryAPI) WithSerial(serialNumber string) *InventoryAPI {
+	c.SerialNumber = serialNumber
+	return c
+}
```
akinross, line 164: Just for my understanding, is the reason for returning on set methods to allow chaining?

internal/manage/api/config_api.go
```diff
@@ -0,0 +1,94 @@
...
+	return c.GetUrl()
+}
+
+func (c *ConfigAPI) DeleteUrl() string {
+	return c.GetUrl()
+}
+
+func (c *ConfigAPI) GetDeleteQP() []string {
+	return nil
+}
```
akinross, line 78: have you considered defining this as default implementations in NDManageAPICommon?

internal/manage/api/api.go
```diff
@@ -0,0 +1,156 @@
...
+	"log"
+	"strings"
+	"sync"
+
+	"github.com/netascode/go-nd"
+
+	"github.com/tidwall/gjson"
+)
+
+type NDManageAPI interface {
```
akinross, line 22: have you considered separating the methods/interfaces further with interface segregation principle for compile time enforcement, or due to nature of go's embedding not preferred?

internal/manage/resource_inventory_switch/debug_dump.go
```diff
@@ -0,0 +1,103 @@
+package resource_inventory_switch
```
akinross, line 9: what is the purpose of this debug_dump file and why is it part of inventory?

internal/manage/resource_fabric_vxlan/fabric_vxlan.go
```diff
@@ -0,0 +1,158 @@
+package resource_fabric_vxlan
```
akinross, line 9: is there some generic rule regarding the file naming?

akinross:
> @mdmohan Would it be okay to separate the vendor removal into its own PR? It would make this PR much quicker to load and review.

Reviewing from commit makes it easier to do

https://github.com/CiscoDevNet/terraform-provider-nd/pull/19/changes/ba567f6a3bccc8aeeb7650f95528f13ae4b1d8e2

generator/defs/fabric.yaml
```diff
@@ -0,0 +1,2831 @@
...
+  - model_name: licenseTier
+    tf_name: license_tier
+    description: The license tier of the fabric_vxlan_evpn resource
+    type: String
+    computed: true
+    optional: true
+    example: premier
+  - model_name: controller
+    tf_name: controller_status
+    ndfc_nested:
```
sajagana, line 21: Do we replace `ndfc` with `nd`? Also for other attributes where we use `ndfc`.

internal/manage/resource_inventory_switch/resource_inventory_switch_test.go
sajagana, line 1: Can you include `_gen` in the auto generated file name?

internal/manage/api/fabric_api.go
```diff
@@ -0,0 +1,103 @@
...
+func (c *FabricAPI) GetUrl() string {
+	// Use the modern API endpoint first if fabric name is specified
+	if c.FabricName != "" {
+		return fmt.Sprintf(UrlFabricByName, c.FabricName)
+	}
+
+	// Fall back to legacy endpoints for special cases
+	if c.Serialnumber != "" {
+		return fmt.Sprintf(urlFabricNameFromSerial, c.Serialnumber)
+	} else if c.FabricName != "" {
```
samiib: Does the logic here together with the if statement above on line 55 work? If FabricName is not an empty string it would have already returned. This case may never be reached. 

Perhaps adding some unit testing would also help.

examples/resources/ndfc_inventory_switch/three_switches.tf
```diff
@@ -0,0 +1,28 @@
...
+  username              = "admin"
+  password              = "mysecret"
+  snmp_v3_auth_protocol = "MD5"
+  max_hop               = 0
+  recalculate           = true
+  deploy                = true
+
+  switches = {
+    "SERIAL001" = {
+      seed_ip         = "10.122.84.63"
```
samiib: Should `seed_ip` here be `ip_address`?

generate.sh
```diff
@@ -0,0 +1,93 @@
...
+$GOPATH/bin/addlicense -c "Cisco Systems, Inc. and its affiliates" -l "mpl" -s  internal/provider/resources/**/*_gen.go
+$GOPATH/bin/addlicense -c "Cisco Systems, Inc. and its affiliates" -l "mpl" -s  internal/provider/ndfc/*_gen.go
+$GOPATH/bin/addlicense -c "Cisco Systems, Inc. and its affiliates" -l "mpl" -s  internal/provider/resources/**/*_test.go
+$GOPATH/bin/addlicense -c "Cisco Systems, Inc. and its affiliates" -l "mpl" -s  internal/provider/provider/**/*_gen.go
+$GOPATH/bin/addlicense -c "Cisco Systems, Inc. and its affiliates" -l "mpl" -s  internal/provider/datasources/**/*_gen.go
+$GOPATH/bin/addlicense -c "Cisco Systems, Inc. and its affiliates" -l "mpl" -s  internal/provider/datasources/**/*_test.go
+$GOPATH/bin/addlicense -c "Cisco Systems, Inc. and its affiliates" -l "mpl" -s  internal/provider/types/
+$GOPATH/bin/addlicense -c "Cisco Systems, Inc. and its affiliates" -l "mpl" -s  internal/provider/*_test.go
+
+$GOPATH/bin/addlicense -c "Cisco Systems, Inc. and its affiliates" -l "mpl" -s  internal/provider/ndfc/*_gen.go
```
samiib: Is this adding licenses to the correct files? I don't think `ndfc` exists here.

generator/defs/nd.yaml
```diff
@@ -0,0 +1,36 @@
+---
+provider:
+  name: nd
+  attributes:
+    - model_name: username
+      tf_name: username
+      type: String
+      mandatory: true
+      description: NDFC Login credentials - user.  Enviroment variable `ND_USERNAME` can be used to override the provider configuration.
```
samiib: typo: Environment
Also looks like an extra space after `user.`

generator/defs/nd.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - model_name: password
+      sensitive: true
+      tf_name: password
+      type: String
+      mandatory: true
+      description: NDFC Login credentials - password. Enviroment variable `ND_PASSWORD` can be used to override the provider configuration.
```
samiib: typo: Environment

generator/defs/nd.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - model_name: domain
+      tf_name: domain
+      type: String
+      mandatory: false
+      description: NDFC Login credentials - domain. Enviroment variable `ND_DOMAIN` can be used to override the provider configuration.
```
samiib: typo: Environment

generator/defs/nd.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - model_name: host
+      tf_name: url
+      type: String
+      mandatory: true
+      description: URL to connect to NDFC - Enviroment variable `ND_URL` can be used to override the provider configuration.
```
samiib: typo: Environment

generator/defs/nd.yaml
```diff
@@ -0,0 +1,36 @@
...
+      tf_name: insecure
+      type: Bool
+      default_value: false
+      description: Controls whether ND server's certificate chain and host name is verified. This can also be set as the `ND_INSECURE` (true or false) environment variable.
+    - model_name: timeout
+      tf_name: timeout
+      type: Int64
+      min_int: 1
+      default_value: 1000
+      description: NDFC HTTP request timeout - timeout. Enviroment variable `ND_TIMEOUT` can be used to override the provider configuration.
```
samiib: typo: Environment

akabhask (dismissed):

docs/index.md
```diff
@@ -1,38 +1,17 @@
...
 description: |-
-  The Cisco ND provider is used to interact with the resources provided by Cisco Nexus Dashboard.
-  The provider needs to be configured with the proper credentials before it can be used.
+  
 ---
 
-# Nexus Dashboard (ND)
+# nd Provider
 
-Cisco Nexus Dashboard is a central management console for multiple data center sites and a common analytics solution for Cisco data center operations. Nexus Dashboard allows users to manage multiple data center sites and provide real time analytics, visibility, assurance for network policies and operations, as well as policy orchestration for the data center fabrics, such as Cisco ACI or Cisco NDFC or standalone Nexus 9000 switches.
```
samiib, line 12: Are there plans to re-add the information that was removed here?

generator/defs/fabric.yaml
```diff
@@ -0,0 +1,2831 @@
...
+    type: String
+    example: eth0
+    description: Telemetry Source Interface (VLAN id or Loopback id) only valid if
+      Telemetry Collection is set to inBand
+  - model_name: telemetrySourceVrf
+    tf_name: telemetry_source_vrf
+    type: String
+    example: vrf1
+    description: VRF over which telemetry is streamed, valid only if telemetry collection
+      is set to inband
```
samiib:
```suggestion
      is set to inBand
```

generator/defs/fabric.yaml
```diff
@@ -0,0 +1,2831 @@
+resource:
+  name: fabric_vxlan
+  generate_tf_resource: true
+  skip_equality: true
+  attributes:
+  - model_name: name
+    tf_name: fabric_name
+    description: The name of the fabric_vxlan_evpn resource
```
samiib: Should we replace `fabric_vxlan_evpn` with just `fabric_vxlan`?

internal/manage/api/config_api.go
samiib: Should these be 2026 now?

internal/provider/provider.go
```diff
@@ -4,223 +4,226 @@ import (
...
+	ctx = tflog.SetField(ctx, "nd_timeout", timeout)
+	ctx = tflog.MaskFieldValuesWithFieldKeys(ctx, "nd_password")
+
+	tflog.Debug(ctx, "Creating Nexus Dashboard client")
+
+	basePath := "/api/v1"
+
+	// Create the shared API client
+	client, err := nd.NewClient(url, basePath, username, password,
+		domain, insecure, nd.MaxRetries(500),
```
samiib: Is max retries hardcoded to 500 here? Is that right? Retries were configurable before but looks like its being removed. Should we keep retries configurable?

internal/provider/provider.go
```diff
@@ -4,223 +4,226 @@ import (
...
+		nd.RequestTimeout(time.Duration(timeout)))
```
samiib:
```suggestion
		nd.RequestTimeout(timeout)
```
Is timeout already time.Duration?

internal/manage/resource_inventory_switch/inventory_switch.go
```diff
@@ -0,0 +1,323 @@
...
+		} else {
+			log.Printf("Switch %s found in fabric %s - not part of resource", sw.SerialNumber, fabricName)
+			continue
+		}
+		outData.Switches[mapKey] = sw.NDFCSwitchesValue
+	}
+
+	input.SetModelData(&outData)
+	DumpInventorySwitchModel("rscGetInventory OUTPUT", input)
+	log.Printf("Read inventory for fabric %s, found %d switches", fabricName, len(outData.Switches))
```
samiib, line 132: Can we replace all the standard logging with tflog instead?

internal/manage/resource_inventory_switch/inventory_switch.go
```diff
@@ -0,0 +1,323 @@
...
+	ModeBootstrap = "bootstrap"
+
+	StatusManageable     = "Manageable"
+	StatusUnreachable    = "Unreachable"
+	StatusPending        = "Pending"
+	StatusFailed         = "Failed"
+	StatusMigrationMode  = "Migration"
+	StatusAlreadyManaged = "alreadyManaged"
+
+	MaxDiscoveryWaitTime = 5 * time.Minute
```
samiib, line 39: Should this wait time be configurable?

samiib (changes requested):

internal/manage/api/fabric_api.go
mdmohan: Not always their own, sometimes shared - so this gives flexibility to select the lock to be used.

Basic goal is to avoid hitting same ND subsystem in parallel.

docs/index.md
mdmohan: Documentation issues shall be worked on separately.
Primary focus would be to get the code working.
There will be issues raised to review/fix documentation

docs/resources/fabric_vxlan.md
mdmohan: As discussed in the sync-up we can track docs issues separately.
Issues can be raised in git to track the same

examples/resources/ndfc_inventory_switch/three_switches.tf
mdmohan: yes; will correct

generator/defs/fabric.yaml
mdmohan: ok, will take care of this

generator/defs/nd.yaml
mdmohan: Will do a detailed doc review (AI assisted) to fix these kind of issues. Thanks for pointing out.

internal/common/types/custtypes.go
mdmohan: custtypes.go is an auto generated file
types could be manual.

internal/manage/api/api.go
mdmohan: QP => query param
rsc is resource

Just individual naming preference.

internal/manage/api/config_api.go
mdmohan: yes, this is copied code from our previous repo.
And this is generated.

We will re-run the generation before making first release.

internal/manage/api/config_api.go
mdmohan: Want the developer to carefully look at all interfaces. 
Providing default will cause misses.
In this case, if something is not explicitly defined, calling it would cause a panic.

internal/manage/api/inventory_api.go
mdmohan: Yes, it allows chaining if needed.

internal/manage/resource_fabric_vxlan/fabric_vxlan.go
mdmohan: Not enforced, but recommendation - I'll add it to Developer Guide.

generator/defs/fabric.yaml
mdmohan: This is internal field, not exposed outside or coming in any documentation. As we are re-using the generator, the field is remaining the same.

internal/manage/resource_inventory_switch/resource_inventory_switch_test.go
mdmohan: _test.go ensures its not compiled inro binary, only built when running go test. putting _test_gen.go will actually compile this code into binary, which is not intended.

internal/manage/api/api.go
mdmohan: Not considered it; 
If it is important, can we do it at a later point after this commit is in?

internal/manage/api/fabric_api.go
mdmohan: Right, the case was from a previous scenario; before the line 55 condition got added; will correct,  its a missed cleanup

generate.sh
mdmohan: errors are ignored, so not a problem; this is copied from previous provider.
I'll correct.

internal/manage/resource_inventory_switch/inventory_switch.go
mdmohan: Had considered in past, 
but the tflog is much noisy, less readable and makes debugging difficult.
log.Printf puts a simple line in the log which can be easily understood.

examples
tflog
```
2026-04-10T18:53:32.159Z [INFO]  provider.terraform-provider-ndfc: Fabric Configure: @caller=/home/ubuntu/go/src/terraform-provider-ndfc/internal/provider/fabric_vxlan_evpn_resource.go:47 tf_provider_addr=registry.terraform.io/cisco/ndfc tf_resource_type=ndfc_fabric_vxlan_evpn tf_rpc=ApplyResourceChange @module=ndfc tf_req_id=277ba4a0-ab76-fa7f-22bd-769f280bbcf3 timestamp=2026-04-10T18:53:32.159Z

```

vs

log.Print

```
026-04-10T18:53:40.689Z [DEBUG] provider.terraform-provider-ndfc: 2026/04/10 18:53:40 [DEBUG] No token available, attempting login...
2026-04-10T18:53:40.689Z [DEBUG] provider.terraform-provider-ndfc: 2026/04/10 18:53:40 [TRACE] Client Login: starting http request
2026-04-10T18:53:41.759Z [DEBUG] provider.terraform-provider-ndfc: 2026/04/10 18:53:41 [DEBUG] Authentication successful
2026-04-10T18:53:41.759Z [DEBUG] provider.terraform-provider-ndfc: 2026/04/10 18:53:41 [DEBUG] HTTP Request: GET, https://172.25.74.175/api/config/dn/apigwcfg/default
2026-04-10T18:53:42.090Z [DEBUG] provider.terraform-provider-ndfc: 2026/04/10 18:53:42 [DEBUG] Exit from doReq method
2026-04-10T18:53:42.090Z [DEBUG] provider.terraform-provider-ndfc: 2026/04/10 18:53:42 [DEBUG] Exit from Do method


```

There are some useful fields in tflog that would help to get the context of any error etc, so I'd use a mix of both, so that we don't miss anything
I'd leave this to the developer, rather than enforcing a rule.

internal/manage/resource_inventory_switch/inventory_switch.go
mdmohan: This resource is a work in progress, have couple more modes to add. I'll consider this aspect after some more testing.

internal/manage/api/api.go
akinross: could be done later, how should we track these enhancements?

internal/manage/api/fabric_api.go
akinross: Why define the mutex at the NDManageAPICommon level? Also here a default implementation would be an option for the GetLock() if in most cases they would be the same.

internal/manage/api/config_api.go
akinross: In that case I would argue that these methods should not be defined at all. Like why implement post, when it is doing get?

internal/manage/api/api.go
akinross: Should we document some form of reference or appendix for these conventions?

docs/resources/fabric_vxlan.md
akinross: Will you raise these with a release blocker tag on it?

internal/common/types/custtypes.go
akinross: and the two other open questions?

would it be helpful to write a comment on the intent of these constructs?

should we use to snakecase for custtypes to keep consistent with other file names something like: cust_types, custom, custom_types ?

internal/manage/resource_fabric_vxlan/fabric_vxlan.go
akinross: thanks, no need to do in this PR as long as we track it!

internal/manage/api/fabric_api.go
mdmohan: The explicit GetLock() per API is intentional — it forces developers to think about which mutex to use for each subsystem. 

A default implementation risks silent misuse where a new API inherits a wrong lock, leading to race conditions that only manifest under concurrent load.

This construct was introduced in the NDFC provider a while back and it  works well and provides flexibility for different use cases — individual locks per resource, shared locks across a group of related APIs, etc.

I'd prefer not to change this at this point and risk introducing a regression/different behaviors. 

If you have an alternate approach that preserves this flexibility, please feel free to enhance it.

internal/manage/api/api.go
mdmohan: I am not sure If I follow this fully. These are just some function names and not an attribute exposed to end user.

There are individual choice of developers - in how a function/method is named

If code readability is your concern, we can introduce some guidelines so that all developers adhere to it.
At the moment we only follow the go style guides.

Feel free to introduce a process/guideline and let everyone follow it.

internal/manage/api/config_api.go
mdmohan: Its the URL to use for the operations, not the operation itself.
  So some resources would have same URL used for POST/GET/PUT where as some different.
  This allows flexibility in defining the URLs clearly, and it has to be explicitly done for each operation.
  Common implementations often causes user to miss the actual point and end up having issues.

internal/provider/provider.go
mdmohan: The retry count is an internal mechanism to ensure we wait long enough before declaring ND unreachable. The user-configurable timeout parameter already controls the overall wait duration. 
Exposing retries adds configuration complexity without clear user benefit — a too-low value would just cause spurious failures. 

Happy to revisit if there's a concrete use case for it.

internal/provider/provider.go
mdmohan: Ya, this is a double cast; I believe its a no-op, so no impacts to intended functionality.
- will change.

internal/manage/api/api.go
mdmohan: Open issues in github

docs/resources/fabric_vxlan.md
mdmohan: Open an issue in github

.github/workflows/checks.yml
mdmohan: Ok, will add it back

internal/manage/api/fabric_api.go
mdmohan: After re-looking at it, I think its just keeping it package private.
Some of those constants may be needed outside the package so they start with caps, and certain others private. Though there is no clear use shown as of now, it might evolve as we develop more resources as other resources might want to use fabric module.

So I'd keep it like that for now. We can cleanup later (changing all to private if not used outside)

go.mod
mdmohan: Ok, so this need to be terraform-provider-nd or else build breaks.
Most of the local imports use relative module path
That is fine as this is a provider and there is no need for doing a <go install github.com/<>>

Will correct this later if needed, as it would need a search and replace of all local imports to use full github.com path.
Note that most of this code was part of a private repo, so the development was done using local imports.

akinross (dismissed):

internal/manage/resource_inventory_switch/debug_dump.go
mdmohan: This is just added for debug. There is no specific reason.

internal/common/types/custtypes.go
mdmohan: need change in generator to change the generated file name. Can do this later.

akabhask (dismissed):

sajagana (dismissed):
LGTM!

internal/provider/provider.go
samiib: @mdmohan Does mean that the client retries 500 times on initial request or on every single request?
Is there a wait time between each request? Is there a condition on when retries occur?
My concern is that this may send too many requests in a short timeframe and cause other gateway/timeout errors. I have seen this happen with newer ND versions.

mdmohan: Retries apply to EVERY individual HTTP request, not just the initial connection.
To prevent flooding of requests there is a back-off mechanism (exponential backoff with jitter) with minimum 4 seconds and max 60 seconds between subsequent retries.

500 now, is just a safe maximum kept for now; We could revise this if we find any better approach.

samiib: We agreed to configure retries and the default is now 100 instead.

samiib (approved):

sajagana (approved):
LGTM!

PR #29, opened 2026-05-07
Common API module and Locking mechanisms

mdmohan:
## Common API Module and Hierarchical Locking

Introduces `internal/common/ndapi/` — a shared API layer with three-level hierarchical locking (Global RWMutex → Fabric RWMutex → Resource Mutex) that replaces per-module lock and HTTP helper duplication.

### Key Changes

- **[locks.go](cci:7://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/locks.go:0:0-0:0)** — [Acquire()](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/locks.go:128:0-177:1) / [LockGuard](cci:2://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/locks.go:106:0-109:1) API with lazy mutex creation; `LockCRUD`, `LockDeploy`, `LockGlobal` modes
- **[ndapi.go](cci:7://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:0:0-0:0)** — [NexusDashboardAPI](cci:2://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:20:0-27:1) interface + [NexusDashboardAPICommon](cci:2://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:29:0-33:1) with auto-locking [Get](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:44:0-59:1)/[Post](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:61:0-92:1)/[Put](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:94:0-127:1)/[Delete](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:129:0-157:1)/[DeployPost](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:159:0-178:1)
- **Migrated all resources** — `fabric_vxlan`, `inventory_switch`, `multi_cluster_connectivity` and deploy workflow
- **Removed duplication** — deleted per-module HTTP wrappers from [manage/api/api.go](cci:7://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/manage/api/api.go:0:0-0:0) and [infra/api/api.go](cci:7://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/infra/api/api.go:0:0-0:0)
- **Resource-agnostic common package** — `ndapi` has zero resource-specific code
- **Deploy safety** — [Post()](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:61:0-92:1)/[Put()](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:94:0-127:1) panic on deploy URLs; [DeployPost()](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:159:0-178:1) acquires no locks (caller holds `LockDeploy`)
- **Updated [Development_Guide.md](cci:7://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/Development_Guide.md:0:0-0:0)** — lock hierarchy, new-resource walkthrough, deploy rules, checklist

### 15 files changed, 804 insertions, 382 deletions


internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, err
+	}
+
+	log.Printf("Finished GET: %s\n", c.NexusDashboardAPI.GetUrl())
+	return res, nil
+}
+
+// Post executes a CRUD POST with hierarchical locking. Payload logging is ON by default;
+// pass disablePayloadLog=true to suppress body logging.
+func (c NexusDashboardAPICommon) Post(payload []byte, disablePayloadLog ...bool) (gjson.Result, error) {
```
akinross: I assume `disablePayloadLog ...bool` is done to avoid changing the call sites? I personally do not think this is the right approach since we would allow multiple bools to be passed in and would be silently ignoring those values. 

Considering the provider is net new we could look into passing a extensible settings like struct.

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+	if len(disablePayloadLog) > 0 {
+		disableLog = disablePayloadLog[0]
+	}
+	if disableLog {
+		mods = append(mods, nd.NoLogPayload)
+	}
+
+	if !json.Valid(payload) {
+		mods = append(mods, nd.RemoveContentType)
+	}
```
akinross: this part seems re-used over various methods, should we create a helper function for this? another option if extensible settings like struct is preferred is a method on these options that constructs the mods based on those settings and payload

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+
+type NexusDashboardAPICommon struct {
+	NexusDashboardAPI
+	Client *nd.Client
+	Fabric string // Fabric scope for locking; defaults to DefaultFabric if empty
+}
+
+// FabricScope returns the fabric name used for lock scoping.
+// Returns DefaultFabric ("global") when Fabric is not set.
+func (c NexusDashboardAPICommon) FabricScope() string {
```
akinross, line 57: why would FabricScope needed here, is this not something that should be enforced at fabric level ( the passed in string? also i see fabricOrDefault() in the locks? are these two related and do we need both?

internal/common/ndapi/locks.go
```diff
@@ -0,0 +1,178 @@
...
+//	Deploy:  globalMu.RLock  → fabricMu[fabric].Lock
+//	Global:  globalMu.Lock
+func Acquire(fabric, resource string, mode LockMode) *LockGuard {
+	fabric = fabricOrDefault(fabric)
+	guard := &LockGuard{}
+
+	switch mode {
+	case LockCRUD:
+		guard.desc = fmt.Sprintf("CRUD %s:%s", fabric, resource)
+		log.Printf("[NDAPI] Acquiring locks: %s\n", guard.desc)
```
akinross, line 144: should we make the log statement more specific to include the file also? 

side question, I noticed log.Printf everywhere is there a reason for not using a dedicated logger (tflog) for logging?

this goes hand-in-hand with thread context.Context and diags frm terraform plugin framework to the functions

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+
+// Post executes a CRUD POST with hierarchical locking. Payload logging is ON by default;
+// pass disablePayloadLog=true to suppress body logging.
+func (c NexusDashboardAPICommon) Post(payload []byte, disablePayloadLog ...bool) (gjson.Result, error) {
+	guard := Acquire(c.FabricScope(), c.NexusDashboardAPI.RscName(), LockCRUD)
+	defer guard.Release()
+
+	url := c.NexusDashboardAPI.PostUrl()
+	if strings.Contains(url, "deploy") {
+		panic("Deploy URL detected in Post call. Use DeployPost method for deployments")
```
akinross, line 85: I think we should leverage typed errors and raise/return them accordingly?

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+	log.Printf("Get URL: %s\n", url)
+	if c.Client == nil {
+		log.Printf("************Client is nil********************")
+	}
+	res, err := c.Client.GetRawJson(url)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Printf("Finished GET: %s\n", c.NexusDashboardAPI.GetUrl())
```
akinross, line 74: cant this be url?
```suggestion
	log.Printf("Finished GET: %s\n", url))
```

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	if disableLog {
+		mods = append(mods, nd.NoLogPayload)
+	}
+
+	if !json.Valid(payload) {
+		mods = append(mods, nd.RemoveContentType)
+	}
+
+	res, err := c.Client.Post(c.NexusDashboardAPI.PostUrl(), string(payload), mods...)
```
akinross: cant this be url?
```suggestion
	res, err := c.Client.Post(url, string(payload), mods...)
```

internal/common/ndapi/ndapi.go
akinross, line 74: I think the logging should be done consistenly across the different methods, a start and end log and the behaviour in between that is worth logging

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	if disableLog {
+		mods = append(mods, nd.NoLogPayload)
+	}
+
+	if !json.Valid(payload) {
+		mods = append(mods, nd.RemoveContentType)
+	}
+
+	res, err := c.Client.Put(c.NexusDashboardAPI.PutUrl(), string(payload), mods...)
```
akinross:
```suggestion
	res, err := c.Client.Put(url, string(payload), mods...)
```

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+
+	qp := c.NexusDashboardAPI.GetDeleteQP()
+	var res nd.Res
+	var err error
+	if qp != nil && body == "" {
+		res, err = c.Client.Delete(c.NexusDashboardAPI.DeleteUrl(), "", func(req *nd.Req) {
+			q := req.HttpReq.URL.Query()
+			for _, s := range qp {
+				keys := strings.Split(s, "=")
+				q.Add(keys[0], keys[1])
```
akinross, line 121: Is it safe to assume the GetDeleteQP always has a `=` else `keys[1]` could lead to a panic, do we want to safeguard with length check?

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+		mods = append(mods, nd.NoLogPayload)
+	}
+
+	if !json.Valid(payload) {
+		mods = append(mods, nd.RemoveContentType)
+	}
+
+	res, err := c.Client.Put(c.NexusDashboardAPI.PutUrl(), string(payload), mods...)
+	if err != nil {
+		return res, err
```
akinross: if the error is not being re-wrapped into something else and nothing happens after why is the conditional needed? could we not simply return res, err directly after the PUT call, same applies to other functions

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+import (
+	"encoding/json"
+	"log"
+	"strings"
+
+	"github.com/netascode/go-nd"
+
+	"github.com/tidwall/gjson"
+)
+
```
akinross, line 20: since the package is already called ndapi, do we need to prepend NexusDashboard to every name?

internal/common/ndapi/ndapi.go
```diff
@@ -0,0 +1,179 @@
...
+type NexusDashboardAPI interface {
+	GetUrl() string
+	PostUrl() string
+	PutUrl() string
+	DeleteUrl() string
```
akinross, line 25: did you consider single URL which takes an operation as attribute?

internal/common/ndapi/ndapi.go
mdmohan: Yes, most of the Post won't need it; only ones having sensitive content, that is inventory

internal/common/ndapi/locks.go
mdmohan: locks are only in this file;

tflog is not very readable, I agree it includes lot of info; but print is much easier to track for quick debugs
So we can use a mix of both tflog and log.Printf.

internal/common/ndapi/ndapi.go
mdmohan: This is not a runtime error; this is intentional crash to prevent someone calling deployment posts outside the preferred framework; reviews may not catch everything, as codebase grows.
Detection and correction expected during UT/IT

internal/common/ndapi/ndapi.go
mdmohan: It should crash so that developer correct it; such mistakes should not go beyond UT/IT. 
If this is caught and logged, no guarantee that someone sees the error.

This is not a user entered string; so not a safety concern IMO.

internal/common/ndapi/ndapi.go
mdmohan: Haven't come across such a usecase; can you give an example?

internal/common/ndapi/ndapi.go
mdmohan: This is just one struct  NexusDashboardAPI; can u clarify what is meant by "every name"?

internal/common/ndapi/ndapi.go
mdmohan: FabricScope() (in ndapi.go) — Public method on NexusDashboardAPICommon. Returns the fabric scope from the API instance's Fabric field. This is what callers and HTTP methods use.
fabricOrDefault() (in locks.go) — Internal safety net inside the lock layer. Ensures Acquire() never receives an empty string, even if someone calls it directly bypassing the API struct.

internal/common/ndapi/ndapi.go
akinross: well NexusDashboardAPICommon also has it, but both ar already in scope of ndapi, so why not simply ACI and Common name?

internal/common/ndapi/ndapi.go
akinross: Does not have to be logged, could also be a specific error being raised to make it more clear even from dev experience

internal/common/ndapi/ndapi.go
akinross: what about his comment: Considering the provider is net new we could look into passing a extensible settings like struct.?

internal/common/ndapi/ndapi.go
akinross: ok this seems like a design issue where we could avoid this redundant check

internal/common/ndapi/ndapi.go
akinross: ok not sure I follow this reasoning, why not expose always and handle error properly? also not sure why this deploy should be avoided in common api

internal/common/ndapi/ndapi.go
akinross: something like this 
```
type Op int

const (
    Get Op = iota
    Post
    Put
    Delete
)

type API interface {
    URL(op Op) string
    ...
    RscName() string
}

File: internal/common/ndapi/ndapi.go (hunk 9 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go:121]: You mean panic? Anyway it crashes; how is the experience different - just a clear string in the panic?
Ok its just a length check, I can add it, but crash/panic is what I will prefer here, not returning some error.

File: internal/common/ndapi/ndapi.go (hunk 5 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go:85]: Deploy needs a global locking to prevent any other ops on fabric.
So special api DeployPost is used; 
This is to prevent anyone attempting a deploy api using normal POST

File: internal/common/ndapi/ndapi.go (hunk 12 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go:25]: Ok, single URL function that returns correct url according to op?

But we have sub conditions
Consider inventory; we use other variables to detect which POST url or which PUT url to use

A single function would become complicated with multiple nested conditions.
Per Operation function splits this and is more readable, IMO. 

So I'd prefer not to change this.

File: internal/common/ndapi/ndapi.go (hunk 11 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go:20]: Ya I considered NDAPI, NDAPICommon, but again 5 uppercase letter didn't look nice. Nor did 'NdAPI'. 

Considering this is cosmetic, I'd keep it  or do a name replace as a seperate commit later after this PR is merged.

File: internal/common/ndapi/ndapi.go (hunk 3 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go:57]: Not sure what is design issue here; pls elaborate

FabricScope() is the public accessor on the API struct; 
It is used to retrieve the Fabric if defined on an API object. It defaults to "global" if not set;

fabricOrDefault() is the lock-layer safety net so Acquire() never sees an empty fabric even if called directly.
lock functions don't carry an object reference, they rely on passed params.

Comment (akinross) [reply internal/common/ndapi/ndapi.go:57]: My point being that if the design would be different we might be able to avoid the exact same function across to different files. 

1. an option could be a shared helper function which both ndapi and lock use
2. another option is designing the api / lock in such a way that the function is not needed in the first place. example would be a fabric struct with a constructor that sets the name to default when not provided in constructor. then making the fabric part of api common.
3. explore other options

Not here to tell you what to design exactly, I am just noting that when I see these duplicated functions across files that typically indicates to me that design could be done differently.

File: internal/common/ndapi/ndapi.go (hunk 5 above)
Comment (akinross) [reply internal/common/ndapi/ndapi.go:85]: But the lock defined in common provides this global locking mechanism? Not sure what the difference is in the post request itself, other then locking constraints which you should be able to provide and path/url? Also not sure when this lock is set correctly how normal posts are then not blocked, is this not the responsibility of the locker to assure this?

File: internal/common/ndapi/ndapi.go (hunk 10 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go]: Ya, I think it was done for logging initially, but there were some duplication and log line got deleted. 
I'll optimize.

File: internal/common/ndapi/ndapi.go (hunk 3 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go:57]: Are you suggesting a struct just to wrap the fabric name? How is a constructor function different from fabricOrDefault, except for readability? Maybe I'm not getting your perspective right.

This IMO is much simpler code, and does what is intended. I'd prefer not to complicate it.
But if you have a concrete design in mind, feel free to submit a PR.

====

As for the duplication, I can change it - `FabricScope()` can delegate to `fabricOrDefault()`, but both entry points are needed since they serve different callers: FabricScope() for the API struct, fabricOrDefault() for direct Acquire() calls.

File: internal/common/ndapi/ndapi.go (hunk 5 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go:85]: Please see [Development_Guide.md](cci:7://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/Development_Guide.md:0:0-0:0) — the lock hierarchy is documented there.

The difference is [Post](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:61:0-88:1) with lock vs [Post](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:61:0-88:1) that doesn't lock (caller already holds lock):

1. **CRUD Post** — takes a resource-level mutex, preventing parallel requests to the same API within a fabric. Sometimes multiple resource types may share the same lock (e.g. all interface resource types).

2. **Deploy Post** — needs a fabric-level **write** lock to block all other Post/Put operations on that fabric. If CRUD operations ran during deploy, recalculate and deployment could fail. This lock is acquired higher up by the caller ([ConfigSaveAndDeploy](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/manage/deployment/deployment.go:60:0-81:1)), which then calls [DeployPost](cci:1://file:///Users/muralidm/code/go/src/terraform-provider-nd/internal/common/ndapi/ndapi.go:130:0-136:1) — a Post that assumes the lock is already held.

3. **Global lock** — freezes all other operations across all fabrics. Not used yet, but the mechanism exists for future features that need it.

This design is carried over from the NDFC provider.
 It's hard to see the full purpose with only one or two resources — it becomes clear as more are added.

File: internal/common/ndapi/ndapi.go (hunk 2 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go]: Ok, will move to a function

File: internal/common/ndapi/ndapi.go (hunk 1 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go]: ok - I'll move it to a struct

File: internal/common/ndapi/ndapi.go (hunk 3 above)
Comment (mdmohan) [reply internal/common/ndapi/ndapi.go:57]: FabricScope duplication addressed as mentioned above
For any other design considerations or way of implementations feel free to refactor.

**Review by akinross — APPROVED** [review]

File: internal/infra/api/api.go [hunk 13]
```diff
@@ -7,140 +7,3 @@
 // SPDX-License-Identifier: MPL-2.0
 
 package api
-
```
akabhask, line 10: Is this file needed?

internal/manage/api/api.go
```diff
@@ -7,150 +7,3 @@
 // SPDX-License-Identifier: MPL-2.0
 
 package api
-
-import (
```
akabhask, line 11: Is this file needed?

akabhask (approved):

akabhask (approved):

internal/manage/api/api.go
mdmohan: Will delete later; As of this PR is serves as a diff

internal/infra/api/api.go
mdmohan: Will delete later; As of this PR is serves as a diff

sajagana (approved):
LGTM!

PR #31, opened 2026-05-13
Fabric vxlan split into vxlan_ibgp and vxlan_ebgp

mdmohan:
Split fabric_vxlan into iBGP/eBGP using fabric_common pattern
Added new eBGP/iBGP resource definitions and generated artifacts.

Old resource fabric_vxlan is retained for reference - will eventually remove once we have more fabric types added.


akabhask (dismissed):

internal/provider/provider.go
```diff
@@ -19,7 +19,6 @@ import (
 	"time"
 
 	_ "terraform-provider-nd/internal/infra/resource_multi_cluster_connectivity"
-	_ "terraform-provider-nd/internal/manage/resource_fabric_vxlan"
 	_ "terraform-provider-nd/internal/manage/resource_inventory_switch"
```
sajagana: We can remove other resource imports right? when we have the register_resource_xyz_gen.go file.

sajagana (changes requested):

internal/provider/provider.go
mdmohan: Yes, that's the intent of register;
I'll remove them

sajagana (approved):
LGTM!

akabhask (approved):

PR #32, opened 2026-05-14
Add nd_local_user resource and test files

### LINKED ISSUE #50: [ND 4.2+] Add nd_local_user resource
### New module/feature description:
```shell
Add Local User resource to support ND 4.2+
```
### Which resource area of Cisco Nexus Dashboard does this new feature belong to?
Infra

sajagana:
```
go test -v -count=1 -timeout 30m ./internal/provider -run TestAccLocalUser
2026/07/07 18:30:52 Test config loaded: url=https://10.15.0.128 fabric=test_fabric_vxlan switches=3 inventory_switches=3
=== RUN   TestAccLocalUserMultiResource
    local_user_resource_test.go:64: Starting testAccPreCheck for global
    local_user_resource_test.go:93: ===== STEP 1: TestAccLocalUserMultiResource_1_create_both_users =====
    local_user_resource_test.go:93: Snapshot: /tmp/tftest_2026_07_07_18-30-52/TestAccLocalUserMultiResource_1_create_both_users.tf
    local_user_resource_test.go:127: ===== STEP 2: TestAccLocalUserMultiResource_2_update_user_one_add_remote_id_claim_and_tenant_domain =====
    local_user_resource_test.go:127: Snapshot: /tmp/tftest_2026_07_07_18-30-52/TestAccLocalUserMultiResource_2_update_user_one_add_remote_id_claim_and_tenant_domain.tf
    local_user_resource_test.go:150: ===== STEP 3: TestAccLocalUserMultiResource_3_import_user_one =====
    local_user_resource_test.go:166: ===== STEP 4: TestAccLocalUserMultiResource_4_import_missing_user =====
--- PASS: TestAccLocalUserMultiResource (13.50s)
=== RUN   TestAccLocalUserResourceCRUD
    local_user_resource_test.go:222: Starting testAccPreCheck for global
    local_user_resource_test.go:247: ===== STEP 1: TestAccLocalUserResourceCRUD_1_create_local_user =====
    local_user_resource_test.go:247: Snapshot: /tmp/tftest_2026_07_07_18-30-52/TestAccLocalUserResourceCRUD_1_create_local_user.tf
    local_user_resource_test.go:274: ===== STEP 2: TestAccLocalUserResourceCRUD_2_modify_scalars =====
    local_user_resource_test.go:274: Snapshot: /tmp/tftest_2026_07_07_18-30-52/TestAccLocalUserResourceCRUD_2_modify_scalars.tf
    local_user_resource_test.go:303: ===== STEP 3: TestAccLocalUserResourceCRUD_3_toggle_xlaunch_extend_roles =====
    local_user_resource_test.go:303: Snapshot: /tmp/tftest_2026_07_07_18-30-52/TestAccLocalUserResourceCRUD_3_toggle_xlaunch_extend_roles.tf
    local_user_resource_test.go:329: ===== STEP 4: TestAccLocalUserResourceCRUD_4_shrink_roles =====
    local_user_resource_test.go:329: Snapshot: /tmp/tftest_2026_07_07_18-30-52/TestAccLocalUserResourceCRUD_4_shrink_roles.tf
--- PASS: TestAccLocalUserResourceCRUD (9.01s)
PASS
ok  	terraform-provider-nd/internal/provider	23.353s
```


internal/infra/resource_local_user/nd_local_user.go
```diff
@@ -0,0 +1,169 @@
...
+			"The input model is nil",
+		)
+		return
+	}
+
+	log.Printf("[INFO] Create nd_local_user login_id=%s", input.LoginId.ValueString())
+	inData := input.GetModelData()
+
+	// Create nd local user API client
+	localUserAPI := api.NewLocalUserAPI(r.infraClient.ApiClient, ndapi.DefaultFabric)
```
akinross: why pass in ndapi.DefaultFabric, should this not default to value? comment in  NexusDashboardAPICommon states: `// Fabric scope for locking; defaults to DefaultFabric if empty`

internal/infra/resource_local_user/local_user_resource_gen.go
```diff
@@ -0,0 +1,506 @@
...
+			"last_name": schema.StringAttribute{
+				Optional:            true,
+				Computed:            true,
+				Description:         "The last name of the local user.",
+				MarkdownDescription: "The last name of the local user.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"login_id": schema.StringAttribute{
```
akinross, line 60: in 4.1.1g UI this is called User ID, how is this in 4.2? is this the correct naming?

internal/infra/resource_local_user/local_user_resource_gen.go
```diff
@@ -0,0 +1,506 @@
...
+				Description:         "The login ID of the local user.",
+				MarkdownDescription: "The login ID of the local user.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"remote_id_claim": schema.StringAttribute{
+				Optional:            true,
+				Computed:            true,
+				Description:         "The CX UserID mapped to the local user.",
```
akinross: what does this description actually mean? Also what does CX stand for in this context?

internal/infra/resource_local_user/local_user_resource_gen.go
```diff
@@ -0,0 +1,506 @@
...
+						ObjectType: types.ObjectType{
+							AttrTypes: SecurityDomainsValue{}.AttributeTypes(ctx),
+						},
+					},
+				},
+				Required:            true,
+				Description:         "The security domains of the local user. At least one security domain must be provided when creating the user.",
+				MarkdownDescription: "The security domains of the local user. At least one security domain must be provided when creating the user.",
+			},
+			"tenant_domain": schema.StringAttribute{
```
akinross, line 110: what is this doing? I do not see this exposed in the https://github.com/CiscoDevNet/ansible-nd/blob/develop/plugins/modules/nd_local_user.py module, is this a new 4.2 feature?

internal/infra/resource_local_user/nd_local_user.go
```diff
@@ -0,0 +1,169 @@
...
+	// Restore reuse_limitation and time_interval_limitation only when they were
+	// known (i.e. user-supplied) prior to the API call, since the API does not
+	// return these fields. If the value was unknown (user did not specify) and
+	// the API did not return it, leave it as Null so Terraform does not see an
+	// unknown value after apply.
+	if !preservedReuseLimitation.IsNull() && !preservedReuseLimitation.IsUnknown() {
+		in.ReuseLimitation = preservedReuseLimitation
+	} else {
+		in.ReuseLimitation = types.Int64Null()
+	}
```
akinross: why do you need the else case:

```
if preservedReuseLimitation.IsNull() || preservedReuseLimitation.IsUnknown() {
    in.ReuseLimitation = types.Int64Null()
}

File: internal/infra/resource_local_user/resource_codec_gen.go [hunk 6]
```diff
@@ -0,0 +1,260 @@
...
+	if jsonData.UserPassword != "" {
+		v.UserPassword = types.StringValue(jsonData.UserPassword)
+	} else {
+		v.UserPassword = types.StringNull()
+	}
+
+	if jsonData.Email != "" {
+		v.Email = types.StringValue(jsonData.Email)
+	} else {
+		v.Email = types.StringNull()
```
akinross, line 60: why are we setting null values for values that can be empty strings? Is it possible to reset a value to empty string once set to a value?

internal/infra/resource_local_user/test_utils_local_user_test.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	for key, value := range c.SecurityDomains {
+		attrNewPath := attrPath.AtName("security_domains").AtName(key)
+		ret = append(ret, SecurityDomainsValueHelperStateCheck(RscName, value, attrNewPath)...)
+	}
+	return ret
+}
+
+func SecurityDomainsValueHelperStateCheck(RscName string, c resource_local_user.NDFCSecurityDomainsValue, attrPath path.Path) []resource.TestCheckFunc {
+	ret := []resource.TestCheckFunc{}
```
samiib: Is this intended to be empty? The same function [here ](https://github.com/CiscoDevNet/terraform-provider-nd/pull/32/changes#diff-bd83495bbc923f0e7d47a2c0ef76bf87ab73977774e3c5593650b5c671f40654R62) contains a check for roles.

internal/infra/resource_local_user/resource_codec_gen.go
```diff
@@ -0,0 +1,260 @@
...
+func (v *SecurityDomainsValue) SetValue(jsonData *NDFCSecurityDomainsValue) diag.Diagnostics {
+
+	var err diag.Diagnostics
+	err = nil
+
+	if len(jsonData.Roles) == 0 {
+		log.Printf("v.Roles is empty")
+		v.Roles = types.SetNull(types.StringType)
+		if err != nil {
+			log.Printf("Error in converting []string to  List %v", err)
```
samiib, line 129: Is this correct? I can't see err being set by SetNull and the log says `converting to List`.

internal/infra/resource_local_user/resource_codec_gen.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	return err
+}
+
+func (v LocalUserModel) GetModelData() *NDFCLocalUserModel {
+	var data = new(NDFCLocalUserModel)
+
+	//MARSHAL_BODY
+
```
samiib, line 151:
```suggestion
```

internal/infra/resource_local_user/resource_codec_gen.go
```diff
@@ -0,0 +1,260 @@
...
+		data.Rbac.TenantDomain = ""
+	}
+
+	if !v.SecurityDomains.IsNull() && !v.SecurityDomains.IsUnknown() {
+		elements1 := make(map[string]SecurityDomainsValue, len(v.SecurityDomains.Elements()))
+
+		data.Rbac.SecurityDomains = make(map[string]NDFCSecurityDomainsValue)
+
+		diag := v.SecurityDomains.ElementsAs(context.Background(), &elements1, false)
+		if diag != nil {
```
samiib, line 207: Is `diag` ever nil? Should this be changed to diag.HasError() like below?

internal/infra/resource_local_user/test_utils_local_user_test.go
```diff
@@ -0,0 +1,59 @@
+// Code generated;  DO NOT EDIT.
+
+package resource_local_user
+
+import (
+	"strconv"
+	"terraform-provider-nd/internal/infra/resource_local_user"
```
samiib: Looks like there are some compile issues in this test file.

samiib (changes requested):

internal/infra/resource_local_user/nd_local_user.go
sajagana: Removed!

internal/infra/resource_local_user/local_user_resource_gen.go
sajagana: Yes, it is User ID in 4.2, changed the attribute name.

internal/infra/resource_local_user/local_user_resource_gen.go
sajagana: Updated

internal/infra/resource_local_user/local_user_resource_gen.go
sajagana: Yes, updated the attribute description.

internal/infra/resource_local_user/nd_local_user.go
sajagana: The local user GET call returns 0 for the ReuseLimitation after the creation in ND 4.2. When the preservedReuseLimitation is not null and not unknown, then need to rewrite the state with preserved value.

internal/infra/resource_local_user/resource_codec_gen.go
sajagana: Need to check with @mdmohan regarding the generated code.

docs/resources/local_user.md
```diff
@@ -38,19 +38,19 @@ resource "nd_local_user" "test_resource_local_user_1" {
...
 - `email` (String) The email address of the local user.
 - `first_name` (String) The first name of the local user.
 - `last_name` (String) The last name of the local user.
-- `remote_id_claim` (String) The CX UserID mapped to the local user.
-- `remote_user_authorization` (Boolean) The user cross launch option of the local user.
+- `remote_id_claim` (String) The Remote ID claim of the local user. This is required when the remote user authorization option is enabled for the local user.
+- `remote_user_authorization` (Boolean) The Remote user authorization is used for signing into Nexus Dashboard when using identity providers that cannot provide authorization claims. Once this attribute is enabled, the local user ID cannot be used to directly login to Nexus Dashboard.
 - `reuse_limitation` (Number) The password reuse limitation of the local user.
-- `tenant_domain` (String) The name of the tenant domain of the local user.
+- `tenant_domain` (String) The name of the tenant domain of the local user. This attribute applies to Nexus Dashboard version 4.2 and later.
```
akinross: is 4.2 not the minimum version of the provider?

akinross (changes requested):

internal/infra/resource_local_user/resource_codec_gen.go
mdmohan: Written out of assumption that we will have either null or valid value.
Is there a real use case for setting to "" ? How is it different from null.

For ND anyway we don't send the attribute if its null or ""

internal/infra/resource_local_user/test_utils_local_user_test.go
sajagana: Please ignore the auto-generated test files.
Valid resource test files are located in the provider directory: `internal/provider/local_user_resource_test.go`

internal/infra/resource_local_user/test_utils_local_user_test.go
sajagana: Please ignore the auto-generated test files.
Valid resource test files are located in the provider directory: `internal/provider/local_user_resource_test.go`

internal/infra/resource_local_user/resource_codec_gen.go
sajagana: In general Set is a list of items, so in the generator it mentioned as a list.

internal/infra/resource_local_user/resource_codec_gen.go
sajagana: This need to be fix in the generator code, let me track it in the issue: https://github.com/ciscoecosystem/nd-codegen/issues/7

internal/infra/resource_local_user/resource_codec_gen.go
sajagana: Issue link: https://github.com/ciscoecosystem/nd-codegen/issues/7

internal/infra/resource_local_user/resource_codec_gen.go
akinross: We discussed this and concluded that for PUT behaviour wise it is not needed for now. I do think that plan output could benefit from the distinction, as null is not the same as empty string. Furthermore if defaults will change to a specific value and a empty string value for some attribute is possible to set to we currently do not allow for this scenario. In the future if we want to leverage PATCH it would also be beneficial to make the distinction between null and empty string. 

Decided to keep as is and change when needed in the future. This might result into breaking behaviour.

internal/infra/resource_local_user/local_user_resource_gen.go
```diff
@@ -0,0 +1,484 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+)
+
+func LocalUserResourceSchema(ctx context.Context) schema.Schema {
+	return schema.Schema{
```
akinross, line 26: Is there no id attribute? I thought this was always a mandatory field to have in terraform schema?

sajagana: As per @mdmohan suggestion, I removed ID from the resource when the object is directly identified by an single attribute. In local_user resource `login_id` is the unique identifier, so `login_id` attribute can be used to import, update and delete the object.

akinross: But typically all terraform resource will have an id defined, why not expose both to keep consistent with other terraform resources? Lets discuss this in next call

akinross:
When a new user is created the user with it's credentials cannot be user, error:  "msg": "{'error': 'Please change temporary password.', 'message': 'Please change temporary password.'}"

Upon UI login when the authentication fails a new UI is shown which forces reset of password top endpoint: https://10.15.0.128/updatepassword

internal/infra/resource_local_user/local_user_resource_gen.go
sajagana: Added id attribute.

docs/resources/local_user.md
```diff
@@ -0,0 +1,61 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "nd_local_user Resource - terraform-provider-nd"
+subcategory: ""
```
akinross, line 4: should we use subcategory? like this would fall under infra bucket if we for instance follow the API / UI buckets

docs/resources/local_user.md
```diff
@@ -0,0 +1,61 @@
...
+description: |-
+  User accounts are provisioned with temporary credentials and require a mandatory password reset at first login to complete onboarding.
+---
+
+# nd_local_user (Resource)
+
+User accounts are provisioned with temporary credentials and require a mandatory password reset at first login to complete onboarding.
```
akinross: Not sure I understand this description, which is more a note. Description to me describes the resource functionality which is basically management of local users. Now what is put in description is also but below. You can leverage ~> I think for notes in md and check that in the preview tool of plugin docs.

docs/resources/local_user.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `login_id` (String) The User ID of the local user.
+- `security_domains` (Attributes Map) The security domains of the local user. At least one security domain must be provided when creating the user. (see [below for nested schema](#nestedatt--security_domains))
+- `user_password` (String, Sensitive) The password of the local user.
```
akinross: here I think the note should be added also, or explain that this is the temporary password of the user which needs to be reset

internal/common/utils/resource.go
```diff
@@ -0,0 +1,110 @@
...
+//
+// Custom usage:
+//
+//	if utils.IsNotFoundError(err, responseBody, utils.NotFoundOptions{
+//		StatusCodes: []int{404},
+//		Messages:    []string{"local user not found"},
+//	}) {
+//		return true
+//	}
+//
```
akinross: why is this here?

internal/common/utils/resource.go
```diff
@@ -0,0 +1,110 @@
...
+
+// SetResourceID copies the resource's unique identifier into the computed
+func SetResourceID(id *types.String, value types.String) string {
+	// Always mirror the resource identifier into Terraform id, including null or unknown.
+	if id != nil {
+		*id = value
+	}
+	if value.IsNull() || value.IsUnknown() {
+		return ""
+	}
```
akinross: when is this ever the cases, should a resource not always have an identifier?

internal/common/utils/resource.go
```diff
@@ -0,0 +1,110 @@
...
+//
+//	if utils.IsNotFoundError(err, responseBody, utils.NotFoundOptions{
+//		StatusCodes: []int{404},
+//		Messages:    []string{"local user not found"},
+//	}) {
+//		return true
+//	}
+//
+// When custom values are supplied, only those custom values are used for matching.
+func IsNotFoundError(err error, response []byte, options ...NotFoundOptions) bool {
```
akinross: could you provide some context on the use case of this function? Is there a deviation in statuscode /. error messages in ND that this is required? the function also appears to do some sanitisation, should that be isolated function with unit tests?

internal/infra/resource_local_user/resource_local_user_test.go
```diff
@@ -0,0 +1,68 @@
...
+	tests := []struct {
+		name string
+		args args
+	}{
+		{
+			name: "Test_resource_local_user",
+			args: args{
+				rscType:  "resource",
+				rscName:  "local_user",
+				dataFile: "/examples/ndfc_payloads/data_local_user.json",
```
akinross, line 28: where is this file located and how is this tested?

sajagana: This file can be used for the unit testing in future.

internal/common/utils/resource.go
sajagana: This function is used to check the not found error from the API response. It can also validate with the custom inputs. When the API does not return 404 status code and proper not found error message. Like when the API returns just null/{} when the object is missing.

Currently I do not have the API endpoint which returns null/{} for the not found error. To avoid this situation I added additional workarounds.

internal/common/utils/resource.go
sajagana: Yes, the unique identifier is always present, removed the function.

internal/common/utils/resource.go
sajagana: This is to give more understanding about the function usage. I can remove if it should not need to be here.

docs/resources/local_user.md
sajagana: Done!

docs/resources/local_user.md
sajagana: Updated the resource doc with new format.

docs/resources/local_user.md
sajagana: Created issue to track the subcategory missing issue: https://github.com/CiscoDevNet/terraform-provider-nd/issues/43

internal/common/utils/resource.go
sajagana: Removed the generic function.

internal/common/utils/resource.go
sajagana: Removed the generic function, this may not be useful for all the cases.

internal/infra/resource_local_user/resource_local_user_test.go
akinross: but why ndfc_payloads?

sajagana: Need to fix the unit test file. For now it is not ready to use.

akinross: is there an issue that tracks this?

sajagana: Created issue to track it: https://github.com/CiscoDevNet/terraform-provider-nd/issues/44

akinross (approved):

internal/infra/resource_local_user/nd_local_user.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		dg.AddError(
+			"Error Creating ND Local User",
+			fmt.Sprintf("Could not create nd local user, Data Marshall error: %v", err),
+		)
+		return
+	}
+
+	// Call the API to create the nd local user
+	res, err := localUserAPI.Post(localUserPayload, nil)
```
akinross: should this disable payload logging with DisablePayloadLog=True set in api options since it contains sensitive values?

enhancement issue could we be to see if we can automatically set this if a schema defines sensitive values which are used in payload.

internal/infra/resource_local_user/nd_local_user.go
```diff
@@ -0,0 +1,176 @@
...
+	// Preserve sensitive fields that are not returned by the API
+	preservedUserPassword := in.UserPassword
+
+	localUserAPI := api.NewLocalUserAPI(r.infraClient.ApiClient)
+	localUserAPI.LoginId = in.LoginId.ValueString()
+	respData, err := localUserAPI.Get()
+
+	if err != nil {
+		if strings.Contains(err.Error(), "StatusCode 404") &&
+			strings.Contains(string(respData), "not found") {
```
akinross: why do we check not Found, and not only the statusCode? also is there no statuscode returned proper which we can check instead of string check?

akinross (changes requested):

internal/infra/resource_local_user/resource_codec_gen.go
```diff
@@ -0,0 +1,231 @@
...
+		for k1, ele1 := range elements1 {
+			data1 := new(NDFCSecurityDomainsValue)
+
+			// roles | List:String| []| false
+			if !ele1.Roles.IsNull() && !ele1.Roles.IsUnknown() {
+
+				listStringData := make([]string, len(ele1.Roles.Elements()))
+				dg := ele1.Roles.ElementsAs(context.Background(), &listStringData, false)
+				if dg.HasError() {
+					panic(dg.Errors())
```
samiib, line 219: Should this propagate errors via return instead of panic?
A panic here will crash the provider process and surface as an error to the user.

samiib (changes requested):

internal/infra/resource_local_user/nd_local_user.go
sajagana: Disabled the payload logging.

internal/infra/resource_local_user/nd_local_user.go
sajagana: Removed additional checks.

internal/infra/resource_local_user/resource_codec_gen.go
sajagana: @mdmohan Are you aware of this issue? Do we need to create TF ND generator issue to track it?

mdmohan: I think it was partially discussed - https://github.com/ciscoecosystem/nd-codegen/issues/7

If error check is missed by caller, then it would result in partial objects causing other inconsistencies.
A crash will make sure any issues with the data structure are fixed during development.

internal/infra/resource_local_user/local_user_resource_gen.go
mdmohan: Pls review the yaml, rather than generated code - Description comes from description field in yaml.

akinross: so we should not review any generated code (assume this is all _gen files)?

internal/infra/resource_local_user/resource_codec_gen.go
sajagana: Created issue to track this issue: https://github.com/ciscoecosystem/nd-codegen/issues/12

akinross (approved):

samiib (approved):

mdmohan (approved):

PR #33, opened 2026-05-17
Update go-nd client usage and add api key support

mtarking:
### Summary

This PR adds support for user-based API key authentication as an alternative to password-based JWT auth, leveraging the updated `go-nd` client library.

#### Provider Schema
- Added `api_key` attribute (optional, sensitive) to the provider configuration
- Changed `password` from required to optional (either `password` or `api_key` must be provided)
- Added `ND_API_KEY` environment variable support

#### Provider Configuration Logic
- Added env var fallback for `ND_API_KEY`
- Updated credential validation: errors only if both `password` and `api_key` are empty
- Passes `nd.UserApiKey(...)` to the `go-nd` client when API key is configured

### Dependencies

This PR requires the updated `go-nd` client ([go-nd PR #23](https://github.com/netascode/go-nd/pull/23)) to be merged first.

### Usage

**Password auth (unchanged):**
```hcl
provider "nd" {
  url      = "https://10.1.1.1"
  username = "admin"
  password = "secret"
}
```

**API key auth (new):**
```hcl
provider "nd" {
  url      = "https://10.1.1.1"
  username = "admin"
  api_key  = "my-api-key"
}
```

**Or via environment variables:**
```bash
export ND_URL="https://10.1.1.1"
export ND_USERNAME="admin"
export ND_API_KEY="my-api-key"
```

### Notes
- Backward compatible — existing configurations with `password` continue to work unchanged
- The `go-nd` update also provides automatic auth endpoint detection (`/api/v1/infra/login` vs `/login`) and token refresh, which work transparently without any provider config changes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module terraform-provider-nd
 
-go 1.24.2
+go 1.25
 
 require (
 	github.com/Jeffail/gabs/v2 v2.7.0
@@ -11,7 +11,7 @@ require (
 	github.com/hashicorp/terraform-plugin-testing v1.14.0
 	github.com/looplab/fsm v1.0.3
 	github.com/netascode/go-nd v0.1.3
-	github.com/stretchr/testify v1.10.0
+	github.com/stretchr/testify v1.11.1
 	github.com/tidwall/gjson v1.18.0
 	golang.org/x/net v0.47.0
 	gopkg.in/yaml.v3 v3.0.1
@@ -70,7 +70,7 @@ require (
 	github.com/posener/complete v1.2.3 // indirect
 	github.com/shopspring/decimal v1.3.1 // indirect
 	github.com/spf13/cast v1.5.0 // indirect
-	github.com/tidwall/match v1.1.1 // indirect
+	github.com/tidwall/match v1.2.0 // indirect
 	github.com/tidwall/pretty v1.2.1 // indirect
 	github.com/tidwall/sjson v1.2.5 // indirect
 	github.com/vmihailenco/msgpack v4.0.4+incompatible // indirect
@@ -93,3 +93,5 @@ require (
 	google.golang.org/protobuf v1.36.9 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/netascode/go-nd => ../../go/go-nd
--- a/internal/provider/nd_provider_gen.go
+++ b/internal/provider/nd_provider_gen.go
@@ -12,6 +12,12 @@ import (
 func NdProviderSchema(ctx context.Context) schema.Schema {
 	return schema.Schema{
 		Attributes: map[string]schema.Attribute{
+			"api_key": schema.StringAttribute{
+				Optional:            true,
+				Sensitive:           true,
+				Description:         "API key for Nexus Dashboard authentication (alternative to password). Environment variable `ND_API_KEY` can be used to override the provider configuration.",
+				MarkdownDescription: "API key for Nexus Dashboard authentication (alternative to password). Environment variable `ND_API_KEY` can be used to override the provider configuration.",
+			},
 			"domain": schema.StringAttribute{
 				Optional:            true,
 				Description:         "NDFC Login credentials - domain. Environment variable `ND_DOMAIN` can be used to override the provider configuration.",
@@ -28,7 +34,7 @@ func NdProviderSchema(ctx context.Context) schema.Schema {
 				MarkdownDescription: "Maximum number of retries for HTTP requests. Environment variable `ND_MAX_RETRIES` can be used to override the provider configuration.",
 			},
 			"password": schema.StringAttribute{
-				Required:            true,
+				Optional:            true,
 				Sensitive:           true,
 				Description:         "NDFC Login credentials - password. Environment variable `ND_PASSWORD` can be used to override the provider configuration.",
 				MarkdownDescription: "NDFC Login credentials - password. Environment variable `ND_PASSWORD` can be used to override the provider configuration.",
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -93,3 +93,5 @@ require (
 	google.golang.org/protobuf v1.36.9 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/netascode/go-nd => ../../go/go-nd
```
mtarking, line 97: To be removed.

internal/provider/provider.go
```diff
@@ -159,12 +164,11 @@ func (p *NexusDashboardProvider) Configure(ctx context.Context, req provider.Con
 		)
 	}
 
-	if config.Password.ValueString() == "" {
-		resp.Diagnostics.AddAttributeError(
-			path.Root("password"),
-			"Missing Nexus Dashboard Password",
-			"The provider cannot create the Nexus Dashboard API client without a password. "+
-				"Set the password value in the configuration or use the ND_PASSWORD environment variable.",
+	if config.Password.ValueString() == "" && config.ApiKey.ValueString() == "" {
```
akinross, line 167: What happens when both are provided? Do we allow both, is there a precedence order and fallback mechanism in place? Else should we restrict providing both inputs together?

internal/provider/provider.go
```diff
@@ -91,6 +91,11 @@ func (p *NexusDashboardProvider) Configure(ctx context.Context, req provider.Con
 		config.Password = types.StringValue(password)
 	}
 
+	if config.ApiKey.IsUnknown() || config.ApiKey.IsNull() {
+		apiKey := os.Getenv("ND_API_KEY")
+		config.ApiKey = types.StringValue(apiKey)
```
akinross, line 96: why assign to a variable when we do not use the variable after?
```suggestion
		config.ApiKey = types.StringValue(os.Getenv("ND_API_KEY"))
```

internal/provider/provider.go
```diff
@@ -191,10 +195,17 @@ func (p *NexusDashboardProvider) Configure(ctx context.Context, req provider.Con
 
 	basePath := "/api/v1"
 
+	// Build client options
+	var mods []func(*nd.Client)
+	mods = append(mods, nd.MaxRetries(int(config.MaxRetries.ValueInt64())))
+	mods = append(mods, nd.RequestTimeout(timeout))
```
akinross, line 201: Should we use a composite literal here?

```suggestion
	mods := []func(*nd.Client){
        nd.MaxRetries(int(config.MaxRetries.ValueInt64())),
        nd.RequestTimeout(timeout),
    }
```

internal/provider/provider.go
```diff
@@ -191,10 +195,17 @@ func (p *NexusDashboardProvider) Configure(ctx context.Context, req provider.Con
...
+	if config.ApiKey.ValueString() != "" {
+		mods = append(mods, nd.UserApiKey(config.ApiKey.ValueString()))
+	}
```
akinross, line 204: We could avoid calling the ValueString() function twice?
```suggestion
	if apiKey := config.ApiKey.ValueString(); apiKey != "" {
        mods = append(mods, nd.UserApiKey(apiKey))
    }
```

internal/provider/provider.go
```diff
@@ -191,10 +195,17 @@ func (p *NexusDashboardProvider) Configure(ctx context.Context, req provider.Con
...
+
 	// Create the shared API client
 	client, err := nd.NewClient(url, basePath, username, password,
-		domain, insecure, nd.MaxRetries(int(config.MaxRetries.ValueInt64())),
-		nd.RequestTimeout(timeout))
+		domain, insecure, mods...)
```
akinross, line 208:
```suggestion
client, err := nd.NewClient(url, basePath, username, password, domain, insecure, mods...)
```

akinross (changes requested):

internal/provider/provider.go
samiib: I think in the client code API key takes precedence and ignores password auth with no fallback.
Making the two options mutually exclusive makes sense. 
Can we use `schemavalidator.AtLeastOneOf`?

internal/provider/nd_provider_gen.go
```diff
@@ -28,7 +34,7 @@ func NdProviderSchema(ctx context.Context) schema.Schema {
 				MarkdownDescription: "Maximum number of retries for HTTP requests. Environment variable `ND_MAX_RETRIES` can be used to override the provider configuration.",
 			},
 			"password": schema.StringAttribute{
-				Required:            true,
+				Optional:            true,
 				Sensitive:           true,
 				Description:         "NDFC Login credentials - password. Environment variable `ND_PASSWORD` can be used to override the provider configuration.",
 				MarkdownDescription: "NDFC Login credentials - password. Environment variable `ND_PASSWORD` can be used to override the provider configuration.",
```
samiib, line 40:
```suggestion
Description:         "ND Login credentials - password. Environment variable `ND_PASSWORD` can be used to override the provider configuration.",
MarkdownDescription: "ND Login credentials - password. Environment variable `ND_PASSWORD` can be used to override the provider configuration.",
```

internal/provider/nd_provider_gen.go
```diff
@@ -12,6 +12,12 @@ import (
 func NdProviderSchema(ctx context.Context) schema.Schema {
 	return schema.Schema{
 		Attributes: map[string]schema.Attribute{
+			"api_key": schema.StringAttribute{
```
mdmohan, line 15: The yaml file update also should be committed or otherwise subsequent re-generates will overwrite these changes - assuming this code was generated. Manual changes are not allowed to _gen.go files.

mdmohan (changes requested):

PR #35, opened 2026-05-25
Add nd_backup resource and test files

sajagana:
```
go test -v -count=1 -timeout 30m ./internal/provider -run TestAccBackup
2026/07/10 21:10:14 Test config loaded: url=https://198.18.133.100 fabric=test_fabric_vxlan switches=3 inventory_switches=3
=== RUN   TestAccBackupResource
    backup_resource_test.go:168: Starting testAccPreCheck for global
    log_step.go:34: ===== STEP 1: TestAccBackupResource_1_create_nd_local_config_only_backup =====
    log_step.go:35: Snapshot: /tmp/tftest_2026_07_21_18-49-09/TestAccBackupResource_1_create_nd_local_config_only_backup.tf
    log_step.go:34: ===== STEP 2: TestAccBackupResource_2_create_nas_full_telemetry_backup =====
    log_step.go:35: Snapshot: /tmp/tftest_2026_07_21_18-49-09/TestAccBackupResource_2_create_nas_full_telemetry_backup.tf
    log_step.go:34: ===== STEP 3: TestAccBackupResource_3_create_scp_config_only_backup =====
    log_step.go:35: Snapshot: /tmp/tftest_2026_07_21_18-49-09/TestAccBackupResource_3_create_scp_config_only_backup.tf
    backup_resource_test.go:159: ===== STEP 4: TestAccBackupResource_4_delete_backup_outside_terraform =====
--- PASS: TestAccBackupResource (701.30s)
=== RUN   TestAccBackupResourceInvalidTelemetryData
    backup_resource_test.go:212: Starting testAccPreCheck for global
    log_step.go:34: ===== STEP 1: TestAccBackupResourceInvalidTelemetryData_1_create_invalid_scp_config_only_telemetry_backup =====
    log_step.go:35: Snapshot: /tmp/tftest_2026_07_21_18-49-09/TestAccBackupResourceInvalidTelemetryData_1_create_invalid_scp_config_only_telemetry_backup.tf
--- PASS: TestAccBackupResourceInvalidTelemetryData (2.44s)
PASS
ok  	terraform-provider-nd/internal/provider	704.522s
```


docs/resources/backup.md
```diff
@@ -0,0 +1,37 @@
...
+### Required
+
+- `destination` (String) The name of the remote location where the backup file will be saved. An empty string indicates a Nexus Dashboard local backup.
+- `encryption_key` (String, Sensitive) The encryption key used to encrypt the backup file. Must be at least 8 characters long and contain at least one letter and one number.
+- `name` (String) The name of the backup. Start with a lowercase letter or digit (a-z, 0-9), end with a lowercase letter or digit, may contain lowercase letters, digits, or hyphens in between, and must be at least 1 character long (a single letter/digit is OK).
+
+### Optional
+
+- `telemetry_data` (Boolean) When set to true, telemetry operational data will be collected. Telemetry operational data can only be collected when `type` is `full` and `destination` is a NAS remote storage location.
+- `type` (String) The type of the backup. `configOnly` backs up the system configuration only, while `full` backs up the system configuration and operational data.
```
akinross: do we document allowed values?

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		dg.AddError(
+			"Error Creating ND Backup",
+			fmt.Sprintf("Could not create nd backup, Data Marshall error: %v", err),
+		)
+		return
+	}
+
+	// Call the API to create the nd backup
+	res, err := backupAPI.Post(backupPayload, nil)
```
akinross: should this disable payload logging with DisablePayloadLog=True set in api options since it contains sensitive values? 

enhancement issue could we be to see if we can automatically set this if a schema defines sensitive values which are used in payload.

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,107 @@
...
+	log.Printf("[INFO] Read nd_backup name=%s", in.Name.ValueString())
+	// Preserve sensitive fields that are not returned by the API
+	preservedEncryptionKey := in.EncryptionKey
+	telemetryData := in.TelemetryData
+
+	backupAPI := api.NewBackupAPI(r.infraClient.ApiClient)
+	backupAPI.Name = in.Name.ValueString()
+	respData, err := backupAPI.Get()
+
+	if err != nil {
```
akinross, line 295: can we manually delete the backup and then have terraform recreate it, or will this lead to errors?

do we currently define test cases for these manual deletion on ND scenarios?

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,107 @@
...
+	// Call the API to create the nd backup
+	res, err := backupAPI.Post(backupPayload, nil)
+	if err != nil {
+		dg.AddError(
+			"Error Creating ND Backup",
+			fmt.Sprintf("Could not create nd backup, unexpected error: %v %v", err, res),
+		)
+		return
+	}
+	r.rscGetBackup(ctx, dg, input)
```
akinross: do we error on all get that fail after rscCreateBackup? could this lead to errors where the state is not saved and the next create tries to recreate, which will error on post because it already exists?

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,107 @@
...
+	in.TelemetryData = telemetryData
+}
+
+// rscDeleteBackup deletes a nd backup by name
+func (r *backupNdResource) rscDeleteBackup(ctx context.Context, dg *diag.Diagnostics, state *BackupModel) {
+	log.Printf("[INFO] Delete nd_backup name=%s", state.Name.ValueString())
+	backupAPI := api.NewBackupAPI(r.infraClient.ApiClient)
+	backupAPI.Name = state.Name.ValueString()
+
+	res, err := backupAPI.Delete(nil)
```
akinross, line 339: what happens when an object does not exist because it was already deleted, does it error?

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// rscCreateBackup creates a nd backup resource
+func (r *backupNdResource) rscCreateBackup(ctx context.Context, dg *diag.Diagnostics, input *BackupModel) {
+	if input == nil {
```
akinross: when is this input ever nil?

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	r.rscGetBackup(ctx, dg, input)
+}
+
+// rscGetBackup retrieves nd backup information by name
+func (r *backupNdResource) rscGetBackup(ctx context.Context, dg *diag.Diagnostics, in *BackupModel) {
+	log.Printf("[INFO] Read nd_backup name=%s", in.Name.ValueString())
+	// Preserve sensitive fields that are not returned by the API
+	preservedEncryptionKey := in.EncryptionKey
+	telemetryData := in.TelemetryData
```
akinross: why is this not called preservedTelemetryData to keep consistent with encryptionKey?

akinross (changes requested):

docs/resources/backup.md
sajagana: Done!

internal/infra/resource_backup/nd_backup.go
sajagana: Done!

internal/infra/resource_backup/nd_backup.go
sajagana: Ignoring the object not found 404 error, to make the delete operation idempotent.

internal/infra/resource_backup/nd_backup.go
sajagana: Could not find the case when the input is nil, removed the condition.

internal/infra/resource_backup/nd_backup.go
sajagana: Done

internal/infra/resource_backup/nd_backup.go
sajagana: Recreating the already existing backup will return an error message returned by the ND.

internal/infra/resource_backup/nd_backup.go
sajagana: Terraform managed object manual deletion case is handled also added the test to validate the manual deletion.

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,253 @@
...
+	"terraform-provider-nd/internal/common/ndapi"
+	"terraform-provider-nd/internal/infra/api"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+const (
+	backupCreatePollInterval = 30 * time.Second
+	backupCreateTimeout      = 60 * time.Minute
```
akinross: should interval/timeout be configurable inputs?

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,253 @@
...
+	backupStatusCompleted    = "completed"
```
akinross: why a constant for a string that is used single time?

internal/infra/resource_backup/nd_backup.go
akinross: did you consider Plugin Framework timeout capability? https://developer.hashicorp.com/terraform/plugin/framework/resources/timeouts

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,253 @@
...
+	id := input.Id.ValueString()
+	backupAPI.Name = id
+	deadline := time.NewTimer(backupCreateTimeout)
+	defer deadline.Stop()
+
+	ticker := time.NewTicker(backupCreatePollInterval)
+	defer ticker.Stop()
+
+	for {
+		respData, err := backupAPI.Get()
```
akinross: should we not reset the timer before every request?

internal/infra/resource_backup/nd_backup.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+}
+
+func (r *backupNdResource) waitForBackupCreateCompletion(ctx context.Context, dg *diag.Diagnostics, backupAPI *api.BackupAPI, input *BackupModel) bool {
+	id := input.Id.ValueString()
+	backupAPI.Name = id
+	deadline := time.NewTimer(backupCreateTimeout)
+	defer deadline.Stop()
+
+	ticker := time.NewTicker(backupCreatePollInterval)
```
akinross: why do we need this ticker exactly is the timer alone not sufficient?

akinross (changes requested):

internal/infra/resource_backup/nd_backup.go
sajagana: Current Schema is generated HashiCorp terraform-plugin-codegen-framework which does not support generating the timeout block: https://github.com/hashicorp/terraform-plugin-codegen-framework/issues/143

internal/infra/resource_backup/nd_backup.go
sajagana: No, we should not reset the timer before every request in waitForBackupCreateCompletion.
Fail if total create wait exceeds timeouts.create / default 90m as per the current change.

internal/infra/resource_backup/nd_backup.go
sajagana: The ticker is still needed to pace the polling of GET /infra/backups/{name} until ND reports status=completed.

internal/infra/resource_backup/nd_backup.go
akinross: @mdmohan ever look at this before? should we add to the generator?

internal/infra/resource_backup/nd_backup.go
akinross: sorry should have made this a bit more clear, that I meant the ticker. because to my understanding of tickers is that are are starting at create time ( line 82 ) and then on a fixed interval would tick. Now we do not know the time it takes to do backupAPI.Get(), but this would be inside the 30 sec window. This window likely becomes smaller ever retry GET we do, possible drop to 0 because when it has already ticked prior to a return of the GET. Would a resettable timer, retry mechanism or a sleep be more suitable for this?

internal/infra/resource_backup/nd_backup.go
akinross: closing this conversation due to: https://github.com/CiscoDevNet/terraform-provider-nd/pull/35#discussion_r3620555464

internal/infra/resource_backup/nd_backup.go
akinross: Besides this we could also explore a generic polling mechanism in the ndapi.go, because likely we will have more of these situation in ND eventually. Where a generic poller can handle timing and cancellation, and endpoint-specific logic (whether 404 is terminal) could be provided with a callback.

internal/infra/resource_backup/nd_backup.go
mdmohan: That would be a nice feature to have, if you think many resources are going to need this timeout feature.
But this is not something in the code we own, but a problem with the Hashicorp's schema generator. 
We may have to fix and contribute, for https://github.com/hashicorp/terraform-plugin-codegen-framework/issues/143,
which I am not sure if anyone is there to accept. 
We will have to diverge to using our clone then.

For now, usage of a variable (as seen here) is a way to go.

We can also check if a normal `Struct` type named `timeouts` (payload hidden) can work as a regular NestedObject - only thing is we won't be able to use the timeouts library in framework, but may give a same usage feel to user

internal/infra/resource_backup/nd_backup.go
sajagana: Thanks for your clarification @akinross , Now I understand the issue and fixed it.

generator/defs/backup.yaml
```diff
@@ -0,0 +1,66 @@
...
+  description: Manages backup for Nexus Dashboard
+  # Documentation note: backup creation waits up to 90 minutes before timing out.
+  # Include this behavior in the generated resource documentation when doc support is updated.
+  attributes:
+  - model_name: id
+    tf_name: id
+    description: The unique identifier of the backup.
+    type: String
+    id: true
+    computed: true
```
mdmohan, line 14: use_state: true needed so that u get it for updates

generator/defs/backup.yaml
```diff
@@ -0,0 +1,66 @@
...
+    mandatory: true
+    tf_requires_replace: true
+    validator: 'LengthBetween(1, 63), stringvalidator.RegexMatches(regexp.MustCompile(`^[a-z0-9]([a-z0-9-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9-]*[a-z0-9])?)*$`), "must start and end with an alphanumeric character and may contain lowercase letters, digits, hyphens, and dots")'
+    example: tf-backup-123
+
+  - model_name: type
+    tf_name: type
+    description: 'The type of the backup. `configOnly` backs up the system configuration only, while `full` backs up the system configuration and operational data. Allowed values: "configOnly" and "full". Default: "configOnly"'
+    type: String
+    optional: true
```
mdmohan: default_value: <> present so no need to mention `optional` here
Everything with default set are optional_computed

internal/common/ndapi/ndapi.go
```diff
@@ -9,15 +9,47 @@
...
 	"github.com/tidwall/gjson"
 )
 
+// ErrPollTimeout is returned when PollUntil reaches its total timeout.
+var ErrPollTimeout = errors.New("poll timed out")
+
+// PollUntil runs poll immediately, then waits interval after each incomplete
+// attempt before retrying. The timeout is a total deadline for the whole poll
+// operation, while interval is the delay between completed poll attempts.
+func PollUntil(ctx context.Context, interval time.Duration, timeout time.Duration, poll func() (bool, error)) error {
```
mdmohan: Looks more like a common utility function to me.
Any reason we should keep under ndapi?

IMO, we should move to a new package like common/util

internal/infra/api/backup_api.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+const RscNameBackup = "backup"
+
+type BackupAPI struct {
+	ndapi.NexusDashboardAPICommon
+	Name string
+}
+
+func NewBackupAPI(client *nd.Client) *BackupAPI {
```
mdmohan: No lock context? 
papi.Fabric is not set - so it will be global b default.
Just making sure its intentional.
Its better to pass lock context name explicitly so that it is clear from code that its not a miss

internal/common/ndapi/ndapi.go
akinross: When making the comment https://github.com/CiscoDevNet/terraform-provider-nd/pull/35#discussion_r3620636662 I was more thinking of extending &ndapi.APIOptions, where we could pass in optional polling mechanism. The polling could be separate package if we want to, do not have a preference for this.

mdmohan: @sajagana atleast move the function to end of file. Not good to have structs and functions mixed up.
Lets add new content to end of the file atleast.

mdmohan: We can think about api-options if similar use cases are seen across other resources

testing/testbed.yaml
```diff
@@ -85,3 +85,11 @@ nd:
       all:
         - "approver"
         - "designer"
+
+  # nd_backup test configuration
+  backup:
```
mdmohan: Don't see anything testbed specific here, can't we just hardcode in test?
The intention of testbed.yml is to have params that change per tested.

generator/defs/backup.yaml
sajagana: Done!

generator/defs/backup.yaml
sajagana: Done!

internal/common/ndapi/ndapi.go
sajagana: For now I just moved the function to end of the file.

internal/infra/api/backup_api.go
sajagana: Done!

testing/testbed.yaml
sajagana: Done!

generator/defs/backup.yaml
```diff
@@ -0,0 +1,88 @@
...
+    computed: true
+    payload_hide: true
+    attributes:
+      - model_name: create
+        tf_name: create
+        description: 'The maximum time to wait for the Nexus Dashboard backup creation process. Default: "90m".'
+        type: String
+        default_value: 90m
+        validator: 'RegexMatches(regexp.MustCompile(`^[1-9][0-9]*m$`), "must be a positive duration in minutes, such as 90m")'
+        tf_requires_replace: true
```
mdmohan, line 82: why would we replace(destroy/create) if this timeout changes?

akinross (dismissed):

internal/infra/resource_backup/backup_resource_gen.go
```diff
@@ -0,0 +1,513 @@
...
+				MarkdownDescription: "The name of the remote location where the backup file will be saved. An empty string indicates a Nexus Dashboard local backup.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"encryption_key": schema.StringAttribute{
+				Required:            true,
+				Sensitive:           true,
+				Description:         "The encryption key used to encrypt the backup file. Must be at least 8 characters long and contain at least one letter and one number.",
+				MarkdownDescription: "The encryption key used to encrypt the backup file. Must be at least 8 characters long and contain at least one letter and one number.",
```
samiib, line 40: Is it possible to add a validator for "at least one letter and one number"? Or are we leaving that validation up to the API?

internal/manage/resource_fabric_common/resource_codec_gen.go
samiib, line 1: Was this file meant to change as part of this PR? Looks a little unrelated.

internal/infra/resource_backup/backup_resource_gen.go
```diff
@@ -0,0 +1,513 @@
...
+				Description:         "The name of the backup. Start with a lowercase letter or digit (a-z, 0-9), end with a lowercase letter or digit, may contain lowercase letters, digits, or hyphens in between, and must be at least 1 character long (a single letter/digit is OK).",
+				MarkdownDescription: "The name of the backup. Start with a lowercase letter or digit (a-z, 0-9), end with a lowercase letter or digit, may contain lowercase letters, digits, or hyphens in between, and must be at least 1 character long (a single letter/digit is OK).",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					stringvalidator.LengthBetween(1, 63), stringvalidator.RegexMatches(regexp.MustCompile(`^[a-z0-9]([a-z0-9-]*[a-z0-9])?(\.[a-z0-9]([a-z0-9-]*[a-z0-9])?)*$`), "must start and end with an alphanumeric character and may contain lowercase letters, digits, hyphens, and dots"),
+				},
+			},
+			"telemetry_data": schema.BoolAttribute{
```
samiib, line 67: Should telemetry_data have a plan modifier if it required replace? In the def is says `tf_requires_replace: true`.

samiib (dismissed):

internal/infra/resource_backup/backup_resource_gen.go
sajagana: Done!

internal/infra/resource_backup/backup_resource_gen.go
sajagana: Done!

internal/manage/resource_fabric_common/resource_codec_gen.go
sajagana: Yes, it is unrelated change to this PR.

generator/defs/backup.yaml
sajagana: All the config file changes should be treated as a force replacement.

mdmohan (approved):

samiib (approved):

PR #39, opened 2026-06-12
Add nd_change_control resource

sajagana:


internal/infra/api/change_control_api.go
```diff
@@ -0,0 +1,61 @@
...
+	papi.Fabric = fabric
+	return papi
+}
+
+func (c *ChangeControlAPI) GetUrl() string {
+	return UrlChangeControl
+}
+
+func (c *ChangeControlAPI) PostUrl() string {
+	// Required by NexusDashboardAPI; the change_control API does not support POST.
```
akinross: if a certain operation is not supported why do we expose the method and return a url?

internal/infra/resource_change_control/nd_change_control.go
```diff
@@ -0,0 +1,100 @@
...
+
+// rscPutChangeControl writes the singleton settings using the API's PUT
+// operation. Create and Update apply the configured settings, while Delete
+// uses the same operation to restore the defaults.
+func (r *changeControlResource) rscPutChangeControl(ctx context.Context, dg *diag.Diagnostics, model *ChangeControlModel) {
+	changeControlAPI := api.NewChangeControlAPI(r.infraClient.ApiClient, ndapi.DefaultFabric)
+
+	if !model.AdminStatus.IsNull() && !model.AdminStatus.IsUnknown() {
+		if !model.Orchestration.IsNull() && !model.Orchestration.IsUnknown() &&
+			!model.NdManagedFabrics.IsNull() && !model.NdManagedFabrics.IsUnknown() {
```
akinross: when are these ever null or unknown when defaults are specified in schema?

internal/infra/resource_change_control/nd_change_control.go
```diff
@@ -0,0 +1,100 @@
...
+			if model.AdminStatus.ValueBool() &&
+				!model.Orchestration.ValueBool() &&
+				!model.NdManagedFabrics.ValueBool() {
+				dg.AddError(
```
akinross: is ND not returning an error for this wrong configuration combination? why handle this in validation in the put instead of something that is seen during plan with something like [ValidateConfig](https://developer.hashicorp.com/terraform/plugin/framework/resources/validate-configuration)?

generator/defs/change_control.yaml
```diff
@@ -0,0 +1,82 @@
...
+    computed: true
+    use_state: true
+    payload_hide: true
+
+  - model_name: changeControlAdminStatus
+    tf_name: admin_status
+    description: The admin status of the change control. The API requires at least one of `orchestration` or `nd_managed_fabrics` to be `true` when this is enabled. Defaults to `false` when not specified in the configuration.
+    type: Bool
+    ndfc_type: bool
+    default_value: false
```
akinross, line 38: Question applies to all attributes, so we enforce the defaults in the provider? are there defaults used by ND when not provided in payload?

docs/resources/change_control.md
```diff
@@ -0,0 +1,54 @@
...
+# nd_change_control (Resource)
+
+Manages change control system settings for Nexus Dashboard.
+
+Change Control is a built-in Nexus Dashboard setting rather than a new object.
+During Terraform creation, import the existing Change Control object first,
+to see differences before applying changes.
+
+During destroy, Terraform uses PUT to restore the default Change Control settings
+before removing the resource from state. The built-in object remains in Nexus Dashboard.
```
akinross, line 27: Should we then also document that if you do not want to have things restored to the terraform defined defaults on destroy that you should delete from state manually?

akinross (changes requested):

internal/infra/api/change_control_api.go
sajagana: Removed the not supported functions

internal/infra/resource_change_control/nd_change_control.go
sajagana: Moved to ValidateConfig function

generator/defs/change_control.yaml
sajagana: Yes, in the TF ND provider we set the default value on the provider side when the ND defaults to some value.

docs/resources/change_control.md
sajagana: Added note to clear the change control state content manually using terraform command.

akinross (approved):

internal/infra/resource_change_control/change_control.go
```diff
@@ -0,0 +1,241 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+}
+
+func (r *changeControlResource) Delete(ctx context.Context, _ resource.DeleteRequest, resp *resource.DeleteResponse) {
+	defaults := ChangeControlModel{
+		Id:                           types.StringValue(changeControlImportID),
```
samiib: Could these defaults be obtained from the generated schema defaults instead? This could be missed if the schema or defaults need to change later.

sajagana: You are correct, thank you @samiib

samiib (approved):
LGTM

akinross (approved):

samiib (approved):
LGTM

docs/resources/change_control.md
```diff
@@ -0,0 +1,63 @@
...
+
+If you want to stop managing Change Control without resetting its current Nexus Dashboard
+settings, remove the resource from Terraform state with `terraform state rm
+nd_change_control.<name>` and remove the resource block from the configuration. This
+bypasses the provider destroy operation and leaves the Nexus Dashboard settings unchanged.
+
+## Example Usage
+
+```terraform
+resource "nd_change_control" "test_resource_change_control_1" {
```
akabhask: Does the example work when tried by someone who starts new? Generally it is good to set values which are default values.

mdmohan (approved):

PR #41, opened 2026-06-15
Add nd_tenant_domain resource

sajagana:


akabhask (dismissed):

docs/resources/tenant_domain.md
```diff
@@ -0,0 +1,37 @@
...
+## Schema
+
+### Required
+
+- `name` (String) The name of the tenant domain.
+
+### Optional
+
+- `description` (String) The description of the tenant domain.
+- `tenant_names` (Set of String) List of tenant name in the tenant domain.
```
samiib: Can tenants be added/removed after creation, or is this a full replacement field? Users may need guidance on whether partial updates are supported, or if the entire tenant set must be respecified on every update.

internal/infra/resource_tenant_domain/nd_tenant_domain.go
```diff
@@ -0,0 +1,165 @@
...
+	id := tenantDomainModel.Id.ValueString()
+	log.Printf("[INFO] Read nd_tenant_domain id=%s", id)
+
+	tenantDomainAPI := api.NewTenantDomainAPI(r.infraClient.ApiClient, ndapi.DefaultFabric)
+	tenantDomainAPI.TenantDomainName = id
+
+	respData, err := tenantDomainAPI.Get()
+	if err != nil {
+		if strings.Contains(err.Error(), "StatusCode 404") {
+			return true
```
samiib: `rscGetTenantDomain` returns `true` when the resource is **not found** and `false` on success, an inverted boolean. 
Would it be better to return an `error` instead? `nil` on success, a sentinel error (e.g., `ErrNotFound`) on 404, or other errors on failure.

internal/infra/resource_tenant_domain/nd_tenant_domain.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		dg.AddError(
+			"Error Updating Tenant Domain",
+			fmt.Sprintf("Could not update tenant domain, unexpected error: %s %s", err.Error(), res.String()),
+		)
+		log.Printf("[ERROR] Error Updating Tenant Domain id=%s: error=%s", id, err.Error())
+		return
+	}
+
+	if r.rscGetTenantDomain(dg, tenantDomainModel) {
```
samiib: I think this is another interesting case were a successful PUT then an unsuccessful GET leaves the desired and actual state out of sync. Is there a way to rollback if GET fails or should it?

docs/resources/tenant_domain.md
sajagana: Tenants can be added to or removed from a tenant domain after creation. The API supports updates through PUT, so changes to the tenant_names attribute trigger an update operation. The attribute represents the complete desired set of tenants associated with the tenant domain.

internal/infra/resource_tenant_domain/nd_tenant_domain.go
sajagana: I like to return the GET call error message and stop the terraform execution. Implementing rollback may be difficult maintain for every resource. This can be discussed with entire team.

samiib (dismissed):
LGTM

samiib (approved):

akabhask (approved):

PR #42, opened 2026-06-24
Fabric external

mdmohan:
Add `nd_fabric_external` resource implementation
- CRUD operations for external fabric
- Acceptance tests


internal/manage/resource_fabric_external/fabric_external.go
```diff
@@ -0,0 +1,164 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Save data into Terraform state
+	resp.Diagnostics.Append(resp.State.Set(ctx, &in)...)
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (r *fabricExternalResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
```
sajagana, line 102: Is the “object not found” error handled?
How is the state cleared or managed when the object is deleted via the UI (without using Terraform)?

internal/manage/resource_fabric_external/fabric_external.go
```diff
@@ -0,0 +1,164 @@
...
+			"The manage module was not registered with the provider.",
+		)
+		return
+	}
+
+	r.manageClient = manageModule.(*manage.NexusDashboardManage)
+}
+
+// Create creates the resource and sets the initial Terraform state.
+func (r *fabricExternalResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
```
sajagana, line 78: Just a question related to the resource lifecycle:

What should we do if the object is already configured when we try to create it with Terraform for the first time?

Should we ignore the “object already exists” error and sync the state with the existing remote object configuration, or should execution stop with an “object already exists” error?

sajagana (changes requested):

internal/manage/resource_fabric_external/fabric_external.go
mdmohan: We have to error out if the resource is already present in infra, to prevent any further damage to existing infra configurations. 

If it exists, user is expected to import it to terraform state.

internal/manage/resource_fabric_external/fabric_external.go
mdmohan: No this is not handled. It errors out if infra is changed outside terraform (UI).
Its not ideal;  will add changed to clear state, so that a drift is detected and plan proposes to add it back.

sajagana: We need to handle the not found error, then only plan proposes to add it back is that correct?

mdmohan: changed - yes Not Found in case of fabric

generator/defs/fabric_common.yaml
```diff
@@ -5,6 +5,12 @@ resource:
   skip_equality: true
   category: manage
   attributes:
+  - model_name: id
+    tf_name: id
+    description: The unique identifier of the fabric_external resource
```
samiib, line 10: Should the common definition reference fabric_external or just fabric? I can see the name below also mentions fabric_vxlan in the common def.

internal/manage/resource_fabric_external/resource_fabric_external_test.go
```diff
@@ -0,0 +1,69 @@
...
+			args: args{
+				rscType:  "resource",
+				rscName:  "fabric_external",
+				dataFile: "/examples/ndfc_payloads/data_fabric_external.json",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name+"model_read", func(t *testing.T) {
+			fileName := os.Getenv("GOPATH") + "/src/terraform-provider-ndfc" + tt.args.dataFile
```
samiib, line 35: Shouldn't this be terraform-provider-nd? I also can't see the example file being committed? 
This test seems to be consistently generated but is it working?

internal/provider/testing/fabric_external.gotmpl
```diff
@@ -0,0 +1,22 @@
+{{define "ND_FABRIC_EXTERNAL_RSC"}}
```
samiib, line 1: There are a lot more fields defined for the resource. Is this set enough coverage?

internal/provider/fabric_external_resource_test.go
```diff
@@ -0,0 +1,185 @@
...
+					FabricExternalModelHelperStateCheck(
+						"nd_fabric_external.fabric_test",
+						*fabricRsc,
+						path.Empty(),
+					)...,
+				),
+			},
+		},
+	})
+}
```
samiib, line 97: Should an Import state step also be added to the tests?

internal/provider/fabric_external_resource_test.go
```diff
@@ -0,0 +1,185 @@
...
+				),
+			},
+			// Step 2: Modify fabric parameters (sub_interface_dot1q_range)
+			{
+				Config: func() string {
+					*stepCount++
+					tName := fmt.Sprintf("%s_%d", t.Name(), *stepCount)
+
+					helper.ModifyFabricExternalObject(&fabricRsc, map[string]interface{}{
+						"sub_interface_dot1q_range": "2-1000",
```
samiib, line 79: Should more updates be tested.
Would it be possible to add an optional field then test it can also be removed?

internal/manage/resource_fabric_external/init.go
```diff
@@ -0,0 +1,19 @@
...
+//
+// SPDX-License-Identifier: MPL-2.0
+
+package resource_fabric_external
+
+import (
+	"terraform-provider-nd/internal/registry"
+)
+
+const ModuleKey = "manage"
```
samiib, line 15: I think fabric_vxlan defines it ModuleKey in fabric_vxlan.go instead of init.go like the rest. Should that be consistently updated?

internal/manage/resource_fabric_vxlan/test_utils_fabric_vxlan_test.go
samiib, line 1: There still appears to be a test that fails to build in fabric_vxlan:

`internal\manage\resource_fabric_vxlan\model_serialization_test.go:33:12: undefined: NDFCFabricVxlanModel`

generator/defs/fabric_common.yaml
mdmohan: This is a  just super set of all fabric fields used for payload; only name and type are relevant for this file; rest can be ignored.
Actual resource schema takes all fields from the their respective files

internal/manage/resource_fabric_external/resource_fabric_external_test.go
mdmohan: This code is not yet used; generated for UT.
We are not running UT now.

internal/provider/fabric_external_resource_test.go
mdmohan: Import is not added into fabric and inventory resources yet, will do as subsequent features

internal/provider/fabric_external_resource_test.go
mdmohan: If you are suggesting more test cases, pls open an issue with the scenario to be tested. Will take it up and implement  separately.

internal/manage/resource_fabric_external/init.go
mdmohan: init.go is the correct place.

fabric_vxlan will be removed at a later point (before release) as the resource functionality was split into vxlan_ibgp and vxlan_ebgp.

As one of the initial resource developed, it is retained for reference.

internal/manage/resource_fabric_vxlan/test_utils_fabric_vxlan_test.go
mdmohan: This is also a file used for some UT case during initial dev.
Not used in the current acceptance tests
Pls ignore for now.

Will be removed when removing resource_fabric_vxlan

internal/provider/testing/fabric_external.gotmpl
mdmohan: Tests will need additional enhancements.
Pls open new issues if some relevant field values combination scenarios are available to be tested

sajagana (approved):
LGTM!

internal/manage/resource_fabric_common/fabric_crud.go
```diff
@@ -80,16 +81,26 @@ func createDefault(ctx context.Context, client *nd.Client, dg *diag.Diagnostics,
 	}
 	// ND BUG - license_tier is not set in the response. Try delaying the read
 	time.Sleep(2 * time.Second)
-	RscGetFabric(ctx, client, dg, model)
```
akinross, line 83: did not review your previous PR but should we statically define a sleeper? should we create a bounded poll that allows us to see if the state is reached?

mdmohan: This is not a state transition problem, but some of the attributes were not reflecting the POST ed value in subsequent GET. Delay was added as a work around, until ND fix the problem.

We had raised an issue with ND, and they had fixed and closed it (CSCwq68525, I didn't verify it)
So ideally this delay is no more needed.
I did not remove it.

Bounded poll may not be needed here.

We can either - 
1) leave it as is as 2 second sleep don't harm much
2) remove it as ND has a fix
Let me know.

akinross: if we do not need it i would say remove or comment the timer, but it should not cause issues. if you think it might cause issues leave it

akinross (approved):

internal/manage/resource_fabric_common/fabric_crud.go
mdmohan: If we ever use an older version without fix, it may fail.
Can we leave it for some time until we have stable 4.2 or 4.3 ND release?

I will need to do some testing if we remove the delay to make sure no other issues manifest.

akinross: please track this as an issue

mdmohan: Done
https://github.com/CiscoDevNet/terraform-provider-nd/issues/46

